PR #45, opened 2015-02-23
Drop VS 2010 and VS 2012 support.

cosmo0920:
Mroonga supports VS 2013 or later.

fixes: https://github.com/groonga/groonga/issues/286


kou:
Please add it to Groonga.

cosmo0920:
I've added it to Groonga: https://github.com/groonga/groonga/pull/287

cosmo0920:
Thank you for merging!!

Issue #50, opened 2015-04-13
Storage mode crashes with same column's ORDER BY


yoku0825:
Mroonga crashes when using ORDER BY two times for same column.
I'll attach test-case soon.

yoku0825:
Simply, we can crash Mroonga by following queries.

```
CREATE TABLE t1 (num int, val varchar(32), PRIMARY KEY(num), FULLTEXT KEY(val)) Engine= Mroonga;
INSERT INTO t1 VALUES (1, 'one');
 
SELECT * FROM t1 WHERE match(val) against('one') ORDER BY num, num LIMIT 1;
```

kou:
Oh...
It seems that "fast order by limit" detection is buggy...

yoku0825:
Oops, I forgot versions which I can reproduce.

Our production enviroment is, 
```
mysql> show global variables like '%version';
+----------------------------+------------+
| Variable_name              | Value      |
+----------------------------+------------+
| innodb_version             | 5.6.23     |
| mroonga_libgroonga_version | 4.1.1      |
| mroonga_version            | 4.10       |
| protocol_version           | 10         |
| version                    | 5.6.23-log |
+----------------------------+------------+
5 rows in set (0.00 sec)
```

And I can reproduce it in Test Environment(latest Groonga and Mroonga) too.
```
mysql56> SHOW GLOBAL VARIABLES LIKE '%version';
+----------------------------+------------+
| Variable_name              | Value      |
+----------------------------+------------+
| innodb_version             | 5.6.24     |
| mroonga_libgroonga_version | 5.0.2      |
| mroonga_version            | 5.01       |
| protocol_version           | 10         |
| version                    | 5.6.24-log |
+----------------------------+------------+
5 rows in set (0.21 sec)
```

kou:
I've fixed it on master. Could you try with nightly build?

yoku0825:
Thank you for your very fast fix!!
I'll try it and inform you a result.

yoku0825:
Oops.. I'm sorry about I forgot to inform you :(
That works fine in our production environment, thanks!

kou:
Thanks for confirming it!

PR #51, opened 2015-04-13
Test case for Issue #50.

yoku0825:
"Storage mode crashes with same column's ORDER BY"


kou:
Thanks. It was helpful.

I created a test case for the problem based on your test case.

yoku0825:
Thank you for your sharing to credit to GMO Media, Inc in the test case.

Issue #69, opened 2015-08-07
mysqld crashes by CREATE TABLE in percona-server-56-mroonga-5.05-1.el6.x86_64.rpm 


yoku0825:
This is caused by #68 .

Official mysql-community-mroonga includes --enable-fast-mutexes in configure option, but percona-server-56-mroonga doesn't so.

## How to repeat.

```
$ sudo yum install -y percona-server-56-mroonga
$ sudo service mysql start
$ mysql -uroot
mysql> CREATE DATABASE d1;
mysql> CREATE TABLE d1.t1 (num int) Engine= Mroonga;
```

## Suggested fix.

a431ffb629076e044582eaa0d022376578f496fe should be apply into percona-server-56-mroonga.

kou:
It has been fixed by your patch.

yoku0825:
Thank you!!

PR #70, opened 2015-08-07
Apply a431ffb to percona-server-56-mroonga.

yoku0825:
This is patch for #69


kou:
It's reasonable. Thanks!

Issue #77, opened 2015-10-29
mysql57-community-mroonga installation fails cause auto-generated password


yoku0825:
Automated install.sql fails by asking password.

https://github.com/mroonga/mroonga/blob/master/packages/rpm/centos/mysql57-community-mroonga.spec.in#L146

## How to repeat

Enter following commands at fresh CentOS 6.6 box.

```
# rpm -ivh http://packages.groonga.org/centos/groonga-release-1.1.0-1.noarch.rpm
# rpm -ivh http://dev.mysql.com/get/mysql57-community-release-el6-7.noarch.rpm
# yum install -y mysql57-community-mroonga
```

And you'll get "Enter password:" prompt (it is in /var/log/mysqld.log)

## Suggested fix

Get auto-generated password in post-script and do *some work* around password expiration..

kou:
Done.

PR #78, opened 2015-10-29
Fix mysql57-community-mroonga's installation issue.

yoku0825:
This is patch for #77 

  - Login and update password using same password string.
    Because of autogenerated password is expired by default.
  - INSTALL PLUGIN using autogenerated password.
  - Expire password for resetting to fresh installed environment.

### DIFF
--- a/packages/rpm/centos/mysql57-community-mroonga.spec.in
+++ b/packages/rpm/centos/mysql57-community-mroonga.spec.in
@@ -24,7 +24,7 @@
 
 Name:		mysql57-community-mroonga
 Version:	@VERSION@
-Release:	1%{?dist}
+Release:	2%{?dist}
 Summary:	A fast fulltext searchable storage engine for MySQL
 
 Group:		Applications/Databases
@@ -107,6 +107,7 @@ rm -rf $RPM_BUILD_ROOT
 if ! /sbin/service mysqld status > /dev/null; then
   /sbin/service mysqld start
   stop_after_installation=1
+  export MYSQL_PWD=`awk '/root@localhost/{print $NF}' /var/log/mysqld.log`
 else
   stop_after_installation=0
 fi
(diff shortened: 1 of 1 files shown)


kou:
Thanks!!!

yoku0825:
@kou I commented at https://github.com/mroonga/mroonga/commit/9f30d4da70b94a43a72debaa1b49760caa0f4eae#commitcomment-14066152

I use MYSQL_PWD with the intention!!

yoku0825:
Because of that, export is necessary too...

kou:
Thanks for your description.
Why did you choose environment variable approach instead of `--password=XXX` approach? Is it for security?

yoku0825:
Because I couldn't determine auto-generated password doesn't have special -- this means which needs to be escaped -- characters for shell, single-quote and double-quote, back-quote.

MYSQL_PWD is free from this concern.

kou:
Thanks. I understood.

(If auto generated password includes single quote or double quote, `"ALTER USER user() IDENTIFIED BY '$MYSQL_PWD'"` will report an error...)

yoku0825:
Oops!? I missed to take care of that statement..
I will look for password generate function tomorrow..

yoku0825:
I found generate password function is here.
https://github.com/mysql/mysql-server/blob/b4104b21520be032400b768cea09a867068be49d/sql/sql_initialize.cc#L81-L144

I seem '!' charset is metachar for bash event but using in variable is safe.

```
$ echo "abc!def12"
-bash: !def12": event not found

$ cat test
abc!def12

$ auto_generated_password=`cat test`
$ echo "$auto_generated_password"
```

Should I fix to use `mysql -p"$auto_generated_password"` ?

kou:
Thanks for confirming it.

> Should I fix to use `mysql -p"$auto_generated_password"` ?

Yes, please! Because it's more explicit code rather than `MYSQL_PWD` environment variable.

yoku0825:
I re-request as #79

PR #79, opened 2015-10-30
Fix to Not using MYSQL_PWD

yoku0825:
Re-fix for #77 .
Discussions are in #78 comments.

### DIFF
--- a/packages/rpm/centos/mysql57-community-mroonga.spec.in
+++ b/packages/rpm/centos/mysql57-community-mroonga.spec.in
@@ -107,7 +107,7 @@ rm -rf $RPM_BUILD_ROOT
 if ! /sbin/service mysqld status > /dev/null; then
   /sbin/service mysqld start
   stop_after_installation=1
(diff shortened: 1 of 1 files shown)


kou:
Thanks!
`eval` is needed?

yoku0825:
No, I just follow line 158.
https://github.com/yoku0825/mroonga/blob/9f001e1cb8044383c15a9796ba3579c89e05e2a8/packages/rpm/centos/mysql57-community-mroonga.spec.in#L158

kou:
I see.
Could you remove `eval`? Because the code is just for reusing command line string for output. In this case, we don't reuse command line string. So we don't need to use `eval`.

kou:
I've merged it because I heard `eval` is needed from @yoku0825.

yoku0825:
I was failing without `eval` but I can determine where causes a problem at last.
I tried `$auto_generated_password` to set like `-p"$auto_generated_password"`, this was mishandled in script.

PR #80, opened 2015-11-02
Fix not using eval

yoku0825:
This refers #79.
At last, I can divide eval and unnecessary variables.


kou:
Nice try!

yoku0825:
I apologized to put you an inconvenience and thank you your merging many times!

PR #89, opened 2015-12-14
Add a new UDF mroonga_normalize()

naoa:
I want to use custom Normalizer of Groonga even if I don't use fulltext index.
So I've implemented mroonga_normalize function as UDF.

```
mroonga_normalize(string, [normalizer_name])
```

```sql
mysql> SELECT mroonga_normalize('aBcＡｂＣ㍑');
+--------------------------------------+
| mroonga_normalize('aBcＡｂＣ㍑') |
+--------------------------------------+
| abcabcリットル                   |
+--------------------------------------+
1 row in set (0.00 sec)
```

It is using NormalizerAuto as default.
And it can be specified any normalizer name as option.

Could you please check and merge it?

### DIFF
--- a/udf/mrn_udf_normalize.cpp
+++ b/udf/mrn_udf_normalize.cpp
@@ -1,0 +1,211 @@
+/* -*- c-basic-offset: 2; indent-tabs-mode: nil -*- */
+/*
+  Copyright(C) 2015 Naoya Murakami <naoya@createfield.com>
+
+  This library is free software; you can redistribute it and/or
+  modify it under the terms of the GNU Lesser General Public
+  License as published by the Free Software Foundation; either
+  version 2.1 of the License, or (at your option) any later version.
+
+  This library is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  Lesser General Public License for more details.
+
+  You should have received a copy of the GNU Lesser General Public
+  License along with this library; if not, write to the Free Software
+  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+#include <mrn_mysql.h>
+#include <mrn_mysql_compat.h>
+#include <mrn_encoding.hpp>
+#include <mrn_windows.hpp>
+#include <mrn_table.hpp>
+#include <mrn_macro.hpp>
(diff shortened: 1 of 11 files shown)


udf/mrn_udf_normalize.cpp
```diff
@@ -0,0 +1,211 @@
...
+              action,
+              info->ctx->errbuf);
+      goto error;
+    }
+  }
+
+  if (args->arg_count == 1) {
+    info->normalizer = grn_ctx_get(info->ctx, DEFAULT_NORMALIZER_NAME, -1);
+  } else {
+    info->normalizer = grn_ctx_get(info->ctx, args->args[1], -1);
```
kou: `args->lengths[1]` should be used.

mysql-test/mroonga/storage/function/normalize/t/japanese.test
```diff
@@ -0,0 +1,40 @@
+# Copyright(C) 2015 Naoya Murakami <naoya@createfield.com>
```
kou: It seems that this file is needless.

naoa:
Thanks for your comment.
(and sorry for wrong commit)
I've fixed them.

udf/mrn_udf_normalize.cpp
```diff
@@ -0,0 +1,211 @@
...
+  }
+  *is_null = 0;
+
+  if (ctx->rc) {
+    my_message(ER_ERROR_ON_WRITE, ctx->errbuf, MYF(0));
+    goto error;
+  }
+
+  *length = result_str->length();
+  DBUG_RETURN((char *)result_str->ptr());
```
kou: You shouldn't use `DBUG_RETURN()` without `MRN_DBUG_ENTER_FUNCTION()` nor `MRN_DBUG_ENTER_METHOD()`.

kou: We need more works in `CMakeLists.txt` and `udf/Makefile.am` to use `DBUG_RETURN()`. (Some build flags should be added.)

But the works aren't the work in this pull request. In other words, you should use `return` instead of `DBUG_RETURN()` in this pull request.

naoa: Thanks.
It has been changed to use return instead of DBUG_RETURN for now.

kou:
Thanks. Could you squash your commits?

naoa:
I've squashed it.

kou:
Thanks!
I've merged this.

naoa:
Thanks for your review and merging!

PR #90, opened 2015-12-15
Omit error messages when checking whether root is password protected or not

relip:
In current version of `mariadb-mroonga.spec`, there is no both stdout and stderr output redirection to /dev/null when running mysql to determine if -p option is needed. Therefore if root account is password protected, the install script prints `ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: NO)`. This kind of error message makes users a bit confusing and to think the installation was unsuccessful even though `Enter password:` prompt follows.

One more thing which didn't patch in this PR is that in the install script, it seems it checks mroonga version (through mysql command) without any password option specified even if `$password_option` is set. This also seems need to be fixed.

### DIFF
--- a/packages/rpm/centos/mariadb-mroonga.spec.in
+++ b/packages/rpm/centos/mariadb-mroonga.spec.in
@@ -86,7 +86,7 @@ mv $RPM_BUILD_ROOT%{_datadir}/doc/mroonga/ mysql-mroonga-doc/
 rm -rf $RPM_BUILD_ROOT
 
 %post
-if /usr/bin/mysql -u root -e "quit"; then
+if /usr/bin/mysql -u root -e "quit" > /dev/null 2>&1; then
   password_option=""
 else
   password_option="-p"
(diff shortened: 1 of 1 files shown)


kou:
Thanks!

> One more thing which didn't patch in this PR is that in the install script, it seems it checks mroonga version (through mysql command) without any password option specified even if `$password_option` is set. This also seems need to be fixed.

Could you send a pull request for the problem?

relip:
The reason I didn't patch is, it can be fixed by several ways and each have its own problem. I'm thinking of two ways, which are 1) Just show the password prompt twice 2) Save the password using `read -s` and pass it to mysql command line directly (`mysql -u root -p$password`). Problems of those two are 1) Users will wonder why same password prompt shows up twice 2) Storing the password into the bash variable might causes (I assume) security issues such as other users on the same server can get the mysql password via `/proc/blah/cmdline`.

kou:
How about using 1) + `--prompt` option of `mysql` command? If we use correct password prompts, users will not be confused.

relip:
Isn't it the option that set the style of the command line that shows up after the login was successful?

kou:
I'm sorry but I couldn't understand what you worry about.
Could you show a concrete example?

PR #96, opened 2016-02-24
storage: convert mysql57's json type to Groonga's vector column

naoa:
I implemented convert mysql57's JSON type field to Groonga's vector column and weight vector column.
This implementation be able to create vector column and weight vector column from MySQL.

How about this implementation?
Currently,  the JSON type doesn't allow any index.
https://github.com/mysql/mysql-server/blob/5.7/sql/sql_table.cc#L4149-L4154
So, It can't be create index to the JSON type field for now...

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -105,6 +105,10 @@
 #  include <create_options.h>
 #endif
 
+#ifdef MRN_HAVE_MYSQL_TYPE_JSON
+#  include <json_dom.h>
+#endif
+
 // for debug
 #define MRN_CLASS_NAME "ha_mroonga"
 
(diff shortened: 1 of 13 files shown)


kou:
Thanks for your pull request.
The idea is acceptable but please wait for a while. I need to release a new version at 2/29...

naoa:
Sure. I am okay anytime. 
I am looking forward to new version.:wave:

kou:
Sorry. I'll handle this pull request in the next month...!

naoa:
必要なユースケースがなくなったので、取り下げます。

kou:
ぜんぜん見れなくてすみませんでした。。。

PR #103, opened 2016-03-29
doc developer release: Use Visual Studio 12(2013) in building instruction in Windows

cosmo0920:
We should use Visual Studio 12(2013) instead of Visual Studio 10(2010) in documents!


kou:
Good catch!

cosmo0920:
Thanks!

PR #137, opened 2017-04-17
cmake: add missing link to libgroonga  when mroonga is bundled and libgroonga isn't bundled

naoa:
It seems there isn't link to libgroonga.so when mroonga is bundled and libgroonga isn't bundled.
Maybe you are not providing this case but I want to bundle mroonga to MySQL without libgroonga for my purpose.


kou:
Thanks!

PR #141, opened 2017-05-21
Support INDEX_MEDIUM/INDEX_SMALL

naoa:
I would like to use ``GRN_OBJ_INDEX_MEDIUM`` flags on Mroonga to optimize memory usage for index of int column.


kou:
Thanks!

PR #143, opened 2017-06-02
add missing null check

kenhys:
When indexes are broken, there is a case that the value of matched_record_keys or
    info->result is equal to null.
    This commit checks these value not to cause unexpected SEGV.

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -8705,6 +8705,14 @@ FT_INFO *ha_mroonga::generic_ft_init_ext(uint flags, uint key_nr, String *key)
     matched_record_keys = grn_table_create(ctx, NULL, 0, NULL,
                                            GRN_OBJ_TABLE_HASH_KEY | GRN_OBJ_WITH_SUBREC,
                                            grn_table, 0);
+    if (!matched_record_keys) {
+      char error_message[MRN_MESSAGE_BUFFER_SIZE];
+      snprintf(error_message, MRN_MESSAGE_BUFFER_SIZE,
+               "failed to create a table for matched record: <%s>",
+               ctx->errbuf);
+      my_message(ER_ERROR_ON_READ, error_message, MYF(0));
+      GRN_LOG(ctx, GRN_LOG_ERROR, "%s", error_message);
+    }
   }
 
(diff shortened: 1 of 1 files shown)


kenhys:
Here is the verification procedure:

1. make patch from this PR.
2. apply patch in spec file
3. install rebuilt mroonga package
4. attach gdb -p to mysqld
5. break ha_mroonga::generic_ft_init_ext
6. mysql> SELECT * FROM diaries WHERE MATCH(content) AGAINST("+fine" IN BOOLEAN MODE); another terminal
7. set matched_record_keys=0 or set info->result=0 in gdb
8. continue! in gdb

This commit works as expected without crash bug (error log is out as follows)

```
2017-06-06 14:41:55.579052|e|66b0d700|/lib64/libc.so.6(clone+0x6d) [0x3f2bce8bbd]
2017-06-06 14:47:02.579880|e|66b0d700|failed to create a table to store result: <>
```

```
$ diff -u mysql57-community-mroonga.spec mysql57-community-mroonga.nullcheck.spec 
--- mysql57-community-mroonga.spec      2017-05-29 14:28:41.000000000 +0900
+++ mysql57-community-mroonga.nullcheck.spec    2017-06-06 14:30:43.580004763 +0900
@@ -32,6 +32,9 @@
 URL:           http://mroonga.org/
 Source0:       http://packages.groonga.org/source/mroonga/mroonga-%{version}.tar.gz
 
+Patch0: 0001-show-error-message-when-creating-a-table-for-matched.patch
+Patch1: 0002-Add-missing-null-check-before-grn_table_setoperation.patch
+
 BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-%(%{__id_u} -n)
 BuildRequires: groonga-devel >= %{groonga_required_version}
 BuildRequires: groonga-normalizer-mysql-devel
@@ -63,6 +66,8 @@
 
 %prep
 %setup -q -n mroonga-%{version}
+%patch0 -p1
+%patch1 -p1
 
 mysql_full_version=%{_mysql_version}-%{_mysql_release}.%{_mysql_dist}
 srpm=mysql-community-${mysql_full_version}.src.rpm
```

kou:
Merged with some modifications.

PR #157, opened 2017-08-02
Fix build requirements for windows

cosmo0920:
To build Mroonga with Visual Studio 2015, it needs to install CMake 3.1 or later.
To build Mroonga with Visual Studio 2017, it needs to install CMake 3.9 or later.

My PowerShell script depends on newly `Invoke-Webrequest` Cmdlet which had been introduced in PowerShell 3.0 or later. (But PowerShell 2.0 is installed in Windows 7 by default.)


kou:
Thanks!

PR #158, opened 2017-08-13
Add mroonga_operations_disable_recording variable

naoa:
I would like to disable recording operations for recovery because It takes a long time to rebuild the index in huge database case.

### DIFF
--- a/doc/source/reference/server_variables.rst
+++ b/doc/source/reference/server_variables.rst
@@ -471,6 +471,33 @@ value.
 
 TODO:
 
+.. _server-variable-mroonga-enable-operations-recording:
+
+``mroonga_enable_operations_recording``
+---------------------------------------
+
+Whether recording operations for recover is enabled or not. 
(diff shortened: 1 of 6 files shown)


doc/source/reference/server_variables.rst
```diff
@@ -471,6 +471,33 @@ value.
 
 TODO:
 
+.. _server-variable-mroonga-operations-disable-recording:
+
+``mroonga_operations_disable_recording``
```
kou: Inverted variable name confuses users.
Please use `mroonga_operations_enable_recording` and the default value is `ON`.

By the way, how about `mroonga_enable_operations_recording`? Because there is already `mroonga_enable_XXX` variable such as `mroonga_enable_optimization`.

naoa:
Thanks for your suggesting.
I've changed  the variable to ``mroonga_enable_operations_recording``.

kou:
Thanks!!

PR #159, opened 2017-08-13
Support generated column

naoa:
`` HA_GENERATED_COLUMNS`` seems to be available just by adding table flags for indexing non virtual(stored) generated column.


kou:
Wow! We just need to update flags to support generated column!
We need more tests for adding/removing generated columns by `ALTER TABLE`.

(We may be able to support MariaDB's computed column by the same way!?)

PR #161, opened 2017-08-13
mariadb 10.2: support virtual column

naoa:
#159 

supported virtual(computed) column of MariaDB.

Virtual(computed) columns starting with 5.2, 10.2.1
https://mariadb.com/kb/en/mariadb/virtual-computed-columns/

JSON functions starting with 10.2.3
https://mariadb.com/kb/en/mariadb/json-functions/

added JSON as an alias for TEXT with 10.2.7
https://jira.mariadb.org/browse/MDEV-9144

PR #162, opened 2017-08-14
Support virtual type without index

naoa:
#159, #161

I've implemented to support ``VIRTUAL`` type for generated/computed column.
It is not supported index.
(MySQL5.7 doesn't allow to create fulltext index for virtual generated column.)


kou:
Thanks!

There are some crashes on Travis CI. Could you confirm them?

naoa:
I've fixed it but I'm not sure why the ``MYSQL_VERSION=10.0.31 MROONGA_TEST_EMBEDDED=yes`` is failed.

kou:
It's caused by timeout. I increased timeout.

PR #163, opened 2017-08-16
Support inplace alter table for stored generated column

naoa:
#161 

I've implemented to support inplace alter table for stored generated column.


kou:
Great!
Can you also add tests?

kou:
Ah, they already exist.

kou:
Thanks!
I've added error handling codes.

naoa:
Thanks for merging!

PR #170, opened 2017-10-05
WIP: Add table_create flag

komainu8:
Todo: 
 * Add test.
 * Administer test.


kou:
`key_large` parameter is too specific.
I've added `flags` parameter that accepts not only `KEY_LARGE` but also `TABLE_HASH_KEY`, `TABLE_PAT_KEY` and `TABLE_DAT_KEY`.

PR #179, opened 2017-12-06
Enable "count-skip" optimization for multi-column fulltext search

fujimotos:
### Summary

Mroonga has [a feature](http://mroonga.org/docs/reference/optimizations.html#row-count) to skip the data retrieval routine when the MySQL/MariaDB server has no interest on the actual row records.

This patch enables the optimimization for a multi-column fulltext index query like this:

```sql
SELECT COUNT(*) FROM tbl WHERE MATCH (col1, col2) AGAINST ('q');
```

### How this patch solves the problem

In essence, this patch eliminates the following `if` conditional in CountSkipChecker:

```cpp
if (select_lex_->select_n_where_fields == 1) {
    ... // Enable optimization
} else {
    ... // Disable optimization
}
```

The observation behind this modification is that this conditional is somewhat meaningless, because, at this point in the code path, the query being executed is guaranteeed to:

* be not a nested query.
* select no field (column) other than `COUNT(*)`
* have no HAVING/GROUP condition.
* involve only a single table.
* have a full-text search func as its root WHERE conditional (`select_lex->where`)

and I cannot come up with a case where we actually require this `if` conditional.

Note that this conditional is introduced by bd4218290. I cannot find further information about this changeset, nor the rationale behind it.

### DIFF
--- a/lib/mrn_count_skip_checker.cpp
+++ b/lib/mrn_count_skip_checker.cpp
@@ -133,18 +133,10 @@ namespace mrn {
       {
         Item_func *func_item = static_cast<Item_func *>(where);
         if (func_item->functype() == Item_func::FT_FUNC) {
-          if (select_lex_->select_n_where_fields == 1) {
-            GRN_LOG(ctx_, GRN_LOG_DEBUG,
-                    "[mroonga][count-skip][true] "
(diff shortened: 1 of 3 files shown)


kou:
Thanks.

PR #182, opened 2017-12-22
package rpm: update Percona Server 5.7

komainu8:
We updated env.sh.in also.

Because added dependence library for percona-server-57-5.7.20-18.1


kou:
Thanks.

PR #217, opened 2018-11-19
mysql8 storage: don't access field name directly in key_part_spec

kenhys:
Since MySQL 8.0.13, field_name of key_part_spec must be accessed by
get_field_name().

TODO: fix can_switch_engines() which is also removed from MySQL 8.0.13

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -4004,13 +4004,12 @@ bool ha_mroonga::storage_create_foreign_key(TABLE *table,
       bool is_same_field_name = false;
       MRN_KEY_PART_SPEC_LIST_EACH_BEGIN(key_spec->columns, key_part_spec) {
         const mrn_key_part_spec_field_name *field_name =
-          &(key_part_spec->field_name);
-        DBUG_PRINT("info", ("mroonga: field_name=%.*s",
-                            static_cast<int>(field_name->length),
-                            field_name->str));
+          MRN_KEY_PART_SPEC_FIELD_NAME(key_part_spec);
+        DBUG_PRINT("info", ("mroonga: field_name=" MRN_KEY_PART_SPEC_FIELD_NAME_FORMAT,
+                            MRN_KEY_PART_SPEC_FIELD_NAME_VALUE(field_name)));
         DBUG_PRINT("info", ("mroonga: field->field_name=" FIELD_NAME_FORMAT,
                             FIELD_NAME_FORMAT_VALUE(field)));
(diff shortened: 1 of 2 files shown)


mrn_mysql_compat.h
```diff
@@ -153,11 +153,40 @@
 #endif
 
 #ifdef MRN_KEY_PART_SPEC_FIELD_NAME_USE_CONST_STRING
+#if (MYSQL_VERSION_ID >= 80013 && !defined(MRN_MARIADB_P))
```
kou: Need 2 spaces between `#` and `if`.

ha_mroonga.cpp
```diff
@@ -4156,7 +4153,11 @@ bool ha_mroonga::storage_create_foreign_key(TABLE *table,
                 "reference column [%s.%s.%s] is not used for primary key",
                 table->s->db.str,
                 normalized_ref_table_name,
+#ifdef MRN_KEY_PART_SPEC_FIELD_NAME_USE_ACCESSSOR_FUNCTION
+                ref_field_name);
+#else
                 ref_field_name->str);
+#endif
```
kou: Can we remove this `ifdef`?

mrn_mysql_compat.h
```diff
@@ -153,11 +153,40 @@
...
+#  define MRN_KEY_PART_SPEC_FIELD_NAME_FORMAT "%.*s"
+#  define MRN_KEY_PART_SPEC_FIELD_NAME_VALUE(key_part_spec)		\
+     static_cast<int>((key_part_spec)->length), (key_part_spec)->str
+#  ifdef MRN_FIELD_FIELD_NAME_IS_LEX_STRING
+#    define MRN_FIELD_NAME_EQUAL_KEY_PART_SPEC_FIELD_NAME(field, string)	\
+       FIELD_NAME_EQUAL_STRING(field, string)
+#  else
+#    define MRN_FIELD_NAME_EQUAL_KEY_PART_SPEC_FIELD_NAME(field, string)	\
+       (strlen((field)->field_name) == string->length &&			\
+       strncmp((field)->field_name, string->str, string->length) == 0)
```
kou: One space is missing.

mrn_mysql_compat.h
kenhys: Thanks! I've fixed.

mrn_mysql_compat.h
kenhys: Oh, I've fixed it.

ha_mroonga.cpp
kenhys: I've just removed.
Thanks.

PR #224, opened 2019-02-07
test: follow changed error message for MariaDB 10.2 or later

kenhys:
Now full name of object is included.

See https://jira.mariadb.org/browse/MDEV-14576


kenhys:
waiting for travis-ci results.

PR #241, opened 2019-05-29
Fix my_byteorder.h related build error

kenhys:


lib/mrn_value_decoder.hpp
```diff
@@ -21,6 +21,7 @@
 #define MRN_VALUE_DECODER_HPP_
 
 #include <mrn_mysql.h>
+#include <mrn_mysql_compat.h>
```
kou: Can we move to this `.cpp`?

lib/mrn_value_decoder.cpp
```diff
@@ -17,9 +17,10 @@
   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+#include <mrn_mysql_compat.h>
```
kou: Including after `mrn_value_decoder.hpp` is better to not affect definitions in `mrn_value_decoder.hpp`.

kenhys: I got it.

kenhys:
@kou could you review again, please?

PR #243, opened 2019-06-11
appveyor: put Mroonga pdb in same directory with plugin

kenhys:
MariaDB search symbols from same directory with plugin.

MariaDB 10.1: pdb is installed to lib/plugin/ha_mroonga.pdb
MariaDB 10.2/10.3: pdb is installed to symbols/ha_mroonga.pdb


kou:
We should do this in `CMakeLists.txt`.
See also: https://github.com/groonga/groonga/blob/master/lib/CMakeLists.txt#L125-L129

PR #245, opened 2019-06-13
appveyor: move install instruction to CMakeLists.txt

kenhys:
It is better way to do so.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -471,3 +471,8 @@ install(FILES
   "${PROJECT_BINARY_DIR}/data/install.sql"
   "${PROJECT_SOURCE_DIR}/data/uninstall.sql"
   DESTINATION "${MRN_DATA_DIR}/")
+if(MSVC AND CMAKE_VERSION VERSION_GREATER "3.1")
+  install(FILES $<TARGET_PDB_FILE:mroonga>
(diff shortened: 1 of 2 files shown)


kenhys:
It works as expected.
https://ci.appveyor.com/project/kenhys/mroonga/builds/25244620

CMakeLists.txt
```diff
@@ -471,3 +471,8 @@ install(FILES
   "${PROJECT_BINARY_DIR}/data/install.sql"
   "${PROJECT_SOURCE_DIR}/data/uninstall.sql"
   DESTINATION "${MRN_DATA_DIR}/")
+if(MSVC AND CMAKE_VERSION VERSION_GREATER "3.1")
+  install(FILES $<TARGET_PDB_FILE:mroonga>
```
komainu8: Where "TARGET_PDB_FILE:mroonga" is defined?
I don't find that definition.

kenhys: @komainu8 

TARGET_PDB_FILE: is CMake builtin feature.
https://cmake.org/cmake/help/v3.1/manual/cmake-generator-expressions.7.html#informational-expressions
So we use `$<TARGET_PDB_FILE:PROJECT_NAME>` is used in here.

komainu8: I see. Thanks!

PR #257, opened 2019-08-07
travis: enable CI for MySQL 8.0 again

kenhys:


kenhys:
@komainu8 This PR is aimed to enable CI for MySQL 8.0 only. I'll create another PR to fix test failures.

tools/travis/before_script.sh
```diff
@@ -70,11 +70,13 @@ else
                       "--with-mysql-build=$PWD/vendor/mysql/release")
       ;;
     mysql-8.0)
-      boost_archive=boost_1_69_0.tar.gz
-      curl -L -O https://dl.bintray.com/boostorg/release/1.69.0/source/${boost_archive}
+      #boost_archive=boost_1_69_0.tar.gz
```
komainu8: Is this comment need?

tools/travis/before_script.sh
```diff
@@ -70,11 +70,13 @@ else
...
+      #curl -L -O https://dl.bintray.com/boostorg/release/1.69.0/source/${boost_archive}
```
komainu8: Ditto.

tools/travis/before_script.sh
```diff
@@ -70,11 +70,13 @@ else
...
       sudo mkdir -p /usr/global/share
-      sudo mv ${boost_archive} /usr/global/share/
+      #sudo mv ${boost_archive} /usr/global/share/
```
komainu8: Ditto.

tools/travis/before_script.sh
kenhys: At first, the download issue will be fixed in upstream, then I think to get them back.(so I've treated them as comments)
Now, it seems better to remove old stuffs. so I'll remove them.

tools/travis/before_script.sh
kenhys: removed.

tools/travis/before_script.sh
kenhys: removed

kenhys:
@komainu8 I've fixed commits. Thanks!

PR #259, opened 2019-08-14
mysql8.0 test: split into build and test jobs

kenhys:
When MROONGA_RELEASE_REQUIREMENT=yes is enabled, at least, minimum
test case must be passed.

### DIFF
--- a/tools/travis/script.sh
+++ b/tools/travis/script.sh
@@ -102,6 +102,39 @@ run_sql_test()
   case "${CI_MYSQL_VERSION}" in
     mysql-8.0)
       n_processors=1
+      # TODO: Remove the following "rm" as soon as possible
+      # when these functionality is supported or test case is fixed for MySQL 8.0.
+      rm -rf ${mroonga_dir}/mysql-test/mroonga/wrapper
+      cd ${mroonga_dir}/mysql-test/mroonga/storage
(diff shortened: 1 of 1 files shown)


kenhys:
These jobs are aimed to check the following:

1.) env mysql-8.0: 
    execute all tests, so it always fails until all tests are fixed for MySQL 8.0
2.) env mysql-8.0 MROONGA_RELEASE_REQUIREMENT=yes 
   execute some tests which are currently supported. so the job will be succeeded anytime)
  (It is useful to check whether  build task is not broken before releasing new one)

kou:
> 1.) env mysql-8.0:
> execute all tests, so it always fails until all tests are fixed for MySQL 8.0

It's not acceptable. Having any always failing job isn't healthy.

kenhys:
So, I'll unify jobs into 2.)

tools/travis/script.sh
```diff
@@ -102,6 +102,38 @@ run_sql_test()
   case "${CI_MYSQL_VERSION}" in
     mysql-8.0)
       n_processors=1
+      # Remove not supported functionality test
```
kou: Could you also add TODO comment to remove these `rm`s as soon as possible?

kenhys: Thanks, I've added comment.

PR #260, opened 2019-08-15
travis appveyor: add MariaDB 10.4.7

kenhys:


appveyor.yml
```diff
@@ -25,7 +25,13 @@ environment:
       MARIADB_VERSION: 10.3.17
       MROONGA_PLATFORM: "winx64"
     - CMAKE_GENERATOR_NAME: "Visual Studio 14 2015 Win64"
-      MARIADB_VERSION: 10.3.17
+      MARIADB_VERSION: 10.4.7
```
kou: This change is invalid.

kou: Ah, my mistake. Sorry.

kou: No.
This should be MariaDB  10.4 + 32bit.

PR #261, opened 2019-08-15
packages yum: add MariaDB 10.4 build task

kenhys:
TODO:  need to fix some build error.
https://travis-ci.org/mroonga/mroonga/jobs/572162590

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -209,7 +209,12 @@ AC_DEFUN([CONFIG_OPTION_MYSQL],[
 
   MYSQL_CXXFLAGS="-felide-constructors"
   case "$MYSQL_MAJOR_MINOR_VERSION" in
-  5.5|5.6|10.*)
+  10.4)
+    m4_ifdef([AX_CXX_COMPILE_STDCXX_11],
+             [AX_CXX_COMPILE_STDCXX_11([ext], [mandatory])],
+             [CHECK_CXXFLAG([-std=gnu++11])])
+    ;;
+  5.5|5.6|10.[0-3])
     MYSQL_CXXFLAGS="$MYSQL_CXXFLAGS -fno-rtti"
     ;;
   8.*)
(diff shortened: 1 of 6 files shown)


packages/yum/build-rpm.sh
```diff
@@ -155,6 +155,24 @@ name = MariaDB
 baseurl = http://yum.mariadb.org/10.3/${distribution}${distribution_version}-${mariadb_architecture}
 gpgkey=https://yum.mariadb.org/RPM-GPG-KEY-MariaDB
 gpgcheck=1
+REPO
+        run yum install -y MariaDB-devel
+        if [ ${distribution_version} = 6 ]; then
+          run yum install -y cmake28
+        fi
+        ;;
+      mariadb-10.4-${PACKAGE})
```
kou: Can we unify `mariadb-10.X-${PACKAGE}` clauses?

configure.ac
```diff
@@ -209,9 +209,14 @@ AC_DEFUN([CONFIG_OPTION_MYSQL],[
 
   MYSQL_CXXFLAGS="-felide-constructors"
   case "$MYSQL_MAJOR_MINOR_VERSION" in
-  5.5|5.6|10.*)
+  5.5|5.6|10.[0-3])
     MYSQL_CXXFLAGS="$MYSQL_CXXFLAGS -fno-rtti"
     ;;
+  10.4)
```
kou: Could you put this before `5.5|5.6|...)`?

packages/rpm/centos/Makefile.am
```diff
@@ -6,6 +6,7 @@ EXTRA_DIST =					\
 	mariadb-10.1-mroonga.spec.in		\
 	mariadb-10.2-mroonga.spec.in		\
 	mariadb-10.3-mroonga.spec.in		\
+	mariadb-10.4-mroonga.spec.in		\
```
kou: This file is missing.

packages/yum/build-rpm.sh
kenhys: I've fixed them for simplicity.

configure.ac
kenhys: I've fixed.

packages/rpm/centos/Makefile.am
kenhys: Oh...
I've added.

PR #268, opened 2019-08-29
appveyor: wait at least 10 seconds for initializing/terminating phase

kenhys:
It avoids not to be initialized/terminated in incomplete state.

### DIFF
--- a/packages/windows/install-mroonga.ps1
+++ b/packages/windows/install-mroonga.ps1
@@ -20,13 +20,18 @@ function Wait-UntilTerminate($cmdName) {
 }
 
 function Install-Mroonga($mariadbVer, $arch, $installSqlDir) {
+  Write-Output("Start to install Mroonga")
   cd "mariadb-$mariadbVer-$arch"
+  Write-Output("Start mysqld.exe")
   Start-Process .\bin\mysqld.exe
-  Wait-UntilRunning mysqld
+  Start-Sleep -s 10
+  Write-Output("Execute install.sql")
   Get-Content "$installSqlDir\install.sql" | .\bin\mysql.exe -uroot
+  Write-Output("Shutdown mysqld.exe")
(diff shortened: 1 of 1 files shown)


packages/windows/install-mroonga.ps1
```diff
@@ -20,13 +20,24 @@ function Wait-UntilTerminate($cmdName) {
...
   cd "mariadb-$mariadbVer-$arch"
+  if ("$mariadbVer" -eq "10.4.7") {
+    Write-Output("Clean data directory")
+    Remove-Item data -Recurse
+    Start-Process .\bin\mysql_install_db.exe
+    Start-Sleep -s 10
+  }
+  Write-Output("Start mysqld.exe")
   Start-Process .\bin\mysqld.exe
-  Wait-UntilRunning mysqld
```
kou: Why do we just run `Start-Sleep -s 10` instead of fixing this function?

kou: We can use `mysqladmin ping` to confirm that `mysqld` is ready: https://github.com/mroonga/docker/blob/master/test/build.sh#L25

packages/windows/install-mroonga.ps1
```diff
@@ -20,13 +20,24 @@ function Wait-UntilTerminate($cmdName) {
...
+  Start-Sleep -s 10
+  Write-Output("Execute install.sql")
   Get-Content "$installSqlDir\install.sql" | .\bin\mysql.exe -uroot
+  Write-Output("Shutdown mysqld.exe")
   Start-Process .\bin\mysqladmin.exe -ArgumentList "-uroot shutdown"
-  Wait-UntilTerminate mysqld
```
kou: Why do we just run `Start-Sleep -s 10` instead of fixing this function?

PR #270, opened 2019-09-02
doc: suppress the warning for "Literal block expected; none found"

komainu8:


doc/source/reference/udf/mroonga_snippet_html.rst
```diff
@@ -20,7 +20,7 @@ highlighted keyword in context.
 Syntax
 ------
 
-``mroonga_snippet_html()`` has required parameter and optional parameter::
+``mroonga_snippet_html()`` has required parameter and optional parameter.
```
kou: Could you add signatures instead of replacing with `.`?

komainu8: What signatures mean?
Is it correct that below modification?

`.` -> `:`

kou:
```rst
``mroonga_snippet_html()`` has required parameter and optional parameter::

  mroonga_snippet_html(document)
```

See also other documents in `doc/source/reference/udf/`.

PR #271, opened 2019-09-04
appveyor: do not use fixed timeout

kenhys:


packages/windows/install-mroonga.ps1
```diff
@@ -4,18 +4,42 @@ Param(
...
-  } while (!$Running)
+    switch ($cmdName) {
+      "mysqld" {
+        if (Test-Path ..\mysqld.txt) {
+          $Version = Get-Content ..\mysqld.txt | Select-String -Pattern "^Version:"
+          if ($Version) {
+            $Waiting = $FALSE
+          }
+        }
+      }
```
kou: This is depends on how to run `mysqld`. So we should not use general name such as `Wait-UntilRunnning`.

We should define `Run-MySQL` instead of using `Start-Process` and `Wait-UntilRunning`.

packages/windows/install-mroonga.ps1
```diff
@@ -4,18 +4,42 @@ Param(
...
+      "mysql_install_db" {
+        if (Test-Path ..\install.txt) {
+          $Successful = Get-Content ..\install.txt | Select-String -Pattern "successful"
+          if ($Successful) {
+            $Waiting = $FALSE
+          }
+        }
+      }
```
kou: ditto.

packages/windows/install-mroonga.ps1
```diff
@@ -4,18 +4,42 @@ Param(
...
   do
   {
-    $Running = Get-Process $cmdName -ErrorAction SilentlyContinue
-    Start-Sleep -m 500
+    if (Test-Path ..\mysqld.txt) {
+      $Complete = Get-Content ..\mysqld.txt | Select-String -Pattern "Shutdown complete"
+      if ($Complete) {
+        $Running = $FALSE
+      }
+    }
```
kou: ditto.

packages/windows/install-mroonga.ps1
kenhys: Thanks, I've fixed to use extracted named function.

packages/windows/install-mroonga.ps1
kenhys: fixed.

packages/windows/install-mroonga.ps1
kenhys: fixed.

PR #277, opened 2019-09-26
test: suppress warning since 10.3.18/10.4.8

kenhys:
Warnings:
  Warning	1280	Name 'property' ignored for PRIMARY key.

### DIFF
--- a/mysql-test/mroonga/storage/index/multiple_column/select/t/varchar.test
+++ b/mysql-test/mroonga/storage/index/multiple_column/select/t/varchar.test
@@ -21,12 +21,14 @@ drop table if exists scores;
 --enable_warnings
 
 set names utf8mb4;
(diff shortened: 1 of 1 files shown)


kou:
We should just remove `property` instead of ignoring warning.

kenhys:
Thanks,　I've fixed it.

kou:
Why do `--disable_warnings` and `--enable_warnings` still exist?

kou:
I think that we need to change `.result` too.

kenhys:
> Why do --disable_warnings and --enable_warnings still exist?

Surely removed at the same time.

https://github.com/mroonga/mroonga/commit/534050c2c8b0656ffce3b9634892d4a9252a2226

kou:
Sorry. I missed it.

PR #283, opened 2019-11-05
packages: support build with docker

kenhys:


kou:
We use different approach.

PR #287, opened 2019-11-19
windows: add a patch for MariaDB 10.4.10

komainu8:
This patch resolves build error.


kou:
Could you use `packages/source/patches/` directory?

PR #302, opened 2020-02-04
packages: add a detection process for updating Ubuntu's packages

komainu8:
If we recompress a source archive when Ubuntu's package update, Launchpad rejects the build request.
Thus, when it updates, we rename the source archive only.


kou:
`rake` doesn't create a new file when it already exists. So we should not do this.

You don't need this change by renaming `mysql-server-5.7-mroonga_9.12.orig.tar.gz` to `mysql-server-5.7-mroonga-9.12.tar.gz`.

PR #310, opened 2020-04-03
doc: use release page instead of direct download links for Windows packages

komainu8:
Because We often forget to update the link,
if we make links that specify the version of MariaDB.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -100,13 +100,13 @@ pygments_style = 'sphinx'
 #modindex_common_prefix = []
 
 variables = {
-  "package_mroonga_version": os.environ["PACKAGE_MROONGA_VERSION"],
-  "download_base": "https://github.com/mroonga/mroonga/releases/tag"
+  "package_mroonga_version": os.environ["PACKAGE_MROONGA_VERSION"]
 }
+
 variables["windows_mariadb_package_link_label"] = \
   "MariaDB with Mroonga-{package_mroonga_version}".format(**variables)
 variables["windows_mariadb_package_link_basename"] = \
-  "{download_base}/v{package_mroonga_version}".format(**variables)
+  "https://github.com/mroonga/mroonga/releases/tag/v{package_mroonga_version}".format(**variables)
 
 rst_epilog = '''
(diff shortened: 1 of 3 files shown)


doc/source/conf.py
```diff
@@ -101,46 +101,16 @@
 
 variables = {
   "package_mroonga_version": os.environ["PACKAGE_MROONGA_VERSION"],
-  "package_mariadb101_version": "10.1.41",
-  "package_mariadb102_version": "10.2.27",
-  "package_mariadb103_version": "10.3.18",
-  "package_mariadb104_version": "10.4.8",
-  "download_base": "https://github.com/mroonga/mroonga/releases/download"
+  "download_base": "https://github.com/mroonga/mroonga/releases/tag"
```
kou, line 104: This is not a base URL for download.

doc/source/conf.py
```diff
@@ -101,46 +101,16 @@
...
-variables["windows_mariadb103_package_link_basename"] = \
-  "{download_base}/v{package_mroonga_version}/mariadb-{package_mariadb103_version}-with-mroonga-{package_mroonga_version}".format(**variables)
-variables["windows_mariadb104_package_link_label"] = \
-  "MariaDB {package_mariadb104_version} with Mroonga {package_mroonga_version}".format(**variables)
-variables["windows_mariadb104_package_link_basename"] = \
-  "{download_base}/v{package_mroonga_version}/mariadb-{package_mariadb104_version}-with-mroonga-{package_mroonga_version}".format(**variables)
+variables["windows_mariadb_package_link_label"] = \
+  "MariaDB with Mroonga-{package_mroonga_version}".format(**variables)
+variables["windows_mariadb_package_link_basename"] = \
+  "{download_base}/v{package_mroonga_version}".format(**variables)
```
kou, line 109: We can expand `{download_base}` to `https://github.com/mroonga/mroonga/releases/tag` here because we have only one `download_base` use.

kou:
Ah, sorry. You filled description in commit message.

doc/source/conf.py
komainu8: I see.
I removed `download_base`.

komainu8:
@kou Thank you for your comments.
I've fixed them.

PR #325, opened 2020-06-05
packages: update patch for mariadb-10.4.13

komainu8:
Because if we use previous the patch, it is occurred build error in CI.


packages/source/patches/mariadb-10.4.13-more-innodb-shutdown-log.diff
```diff
@@ -0,0 +1,167 @@
+--- mariadb-10.4.13.orig/storage/innobase/srv/srv0start.cc	2020-05-10 04:28:02.000000000 +0900
++++ mariadb-10.4.13/storage/innobase/srv/srv0start.cc	2020-06-05 15:26:09.621468003 +0900
+@@ -2426,6 +2426,7 @@
+ /** Shut down InnoDB. */
+ void innodb_shutdown()
+ {
++        ib::info() << "Shutdown started; operation " << srv_operation;
```
kou, line 7: It seems that indent is broken. (Tab characters are expanded.)

PR #336, opened 2020-08-18
test: follow error message change in mariadb-10.5.3, 10.4.13, 10.3.23, and 10.2.32

komainu8:
Because the format of the error message in mariadb-server was modified by below commit.

https://github.com/MariaDB/server/commit/cb4da5da74b7a6f2e7c4f4ed1b0e5affe45fe2a2

However, the format of it is not modified in MySQL, PerconaServer, and MariaDB 10.1.
Therefore, we separated `mroonga/storage/index/multiple_column/unique/{t, r}/decimal.{test, result}`.

### DIFF
--- a/mysql-test/mroonga/storage/index/multiple_column/unique/t/decimal.test
+++ b/mysql-test/mroonga/storage/index/multiple_column/unique/t/decimal.test
@@ -30,7 +30,9 @@ select c1, c2, c3 from t1 force index(uk1) where c2 = -9876543210987654321098765
 select c1, c2, c3 from t1 force index(uk1) order by c2, c3;
 select c1, c2, c3 from t1 force index(uk1) order by c2 desc, c3 desc;
 select c2, c3 from t1 force index(uk1) order by c2, c3;
-#For MariaDB 10.2 or later
+# MariaDB 10.2 or later uses "..." to show the displayed value is truncated.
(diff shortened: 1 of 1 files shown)


kou:
Can we use `--replace_result` instead of duplicate the test like https://github.com/mroonga/mroonga/blob/master/mysql-test/mroonga/storage/repair_table/t/no_index_file.test#L43 ?

komainu8:
@kou Thank you for your comment.
I've fixed it.

mysql-test/mroonga/storage/index/multiple_column/unique/t/decimal.test
```diff
@@ -30,6 +30,8 @@ select c1, c2, c3 from t1 force index(uk1) where c2 = -9876543210987654321098765
 select c1, c2, c3 from t1 force index(uk1) order by c2, c3;
 select c1, c2, c3 from t1 force index(uk1) order by c2 desc, c3 desc;
 select c2, c3 from t1 force index(uk1) order by c2, c3;
+#For MariaDB 10.2 or later
```
kou, line 33:
```suggestion
# MariaDB 10.2 or later uses "..." to show the displayed value is truncated.
# We will use "..." version instead of "000" version as expected value
# when MySQL also uses "...".
```

PR #348, opened 2020-09-11
ci: add a new job for checking installation on Ubuntu

komainu8:


komainu8:
Mroonga 11.02のUbuntu向けのパッケージのインストールに失敗する。

```
$ sudo apt install -y -V mariadb-server-mroonga
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 mariadb-server-mroonga : Depends: mariadb-server-10.3-mroonga (= 11.02-1.ubuntu20.04.1) but it is not going to be installed
                          Depends: mariadb-server (= 1:10.3.25-0ubuntu0.20.04.1) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

komainu8:
`mariadb-server (= 1:10.3.25-0ubuntu0.20.04.1)`を要求しているが、Ubuntu 20.04のmariadb-serverは、`10.3.29-0ubuntu0.20.04.1`なので、依存パッケージのバージョンの指定がおかしい。

komainu8:
`mariadb-serve`のバージョンを指定しているのは、`packages/mariadb-server-10.3-mroonga/debina/control.in`
この中で `mariadb-server-10.3 (= @MYSQL_VERSION@)` と定義されているので、`@MYSQL_VERSION@`の値を入れているところを直せば良さそう。

.github/workflows/install.yml
```diff
@@ -0,0 +1,40 @@
+name: Install to Ubuntu
+on:
+  push:
+    branches:
+      - "*"
+    tags:
+      - "*"
```
kou, line 7: We don't need to run this on `push`:

```suggestion
```

.github/workflows/install.yml
```diff
@@ -0,0 +1,40 @@
...
+        0 0 * * *
+jobs:
+  install:
+    name: Install to Ubuntu
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-18.04
+            package-type: apt
```
kou, line 19: Do we need this?
It seems that we can always use `apt`.

packages/apt/install_to_ubuntu.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -exu
+
+package=$1
+
+sudo apt update
+sudo apt-get install -y -V software-properties-common lsb-release
```
kou, line 8: Could you use `apt` instead of `apt-get`?

.github/workflows/install.yml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: actions/checkout@v2
+      - name: Run VM
+        run: |
+          vagrant up ${{ matrix.os }}
+      - name: Install Mroonga
+        run: |
+          vagrant \
+            ssh ${{ matrix.os }} \
+            -- \
+            /vagrant/packages/${{ matrix.package-type }}/install_to_ubuntu.sh \
```
kou, line 38: How about `install_test.sh` instead of `install_to_ubuntu.sh`?

kou (changes requested):

.github/workflows/install.yml
komainu8: Ah... I see.
I've fixed it.

.github/workflows/install.yml
komainu8: Oh... This is needless.
I've fixed it.

.github/workflows/install.yml
komainu8: I agree with you.
I've fixed it.

packages/apt/install_to_ubuntu.sh
komainu8: Ah...
I've fixed it.

komainu8:
@kou Thank you for your comments.
I've fixed them. Could you review them again?

.github/workflows/install.yml
kou:
```suggestion
            /vagrant/packages/apt/install_test.sh \
```

.github/workflows/install.yml
```diff
@@ -0,0 +1,31 @@
...
+        0 0 * * *
+jobs:
+  install:
+    name: Install to Ubuntu
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-18.04
+            package: mysql-server
```
kou, line 14: We'll change package name by the next release:

```suggestion
            package: mysql-5.7
```

.github/workflows/install.yml
```diff
@@ -0,0 +1,31 @@
...
+          - os: ubuntu-20.04
+            package: mariadb-server
```
kou, line 16: We'll change package name by the next release:

```suggestion
            package: mariadb
```

.github/workflows/install.yml
```diff
@@ -0,0 +1,31 @@
...
+          - os: ubuntu-20.10
+            package: mariadb-server
```
kou, line 18: We'll change package name by the next release:

```suggestion
            package: mariadb
```

komainu8:
I've fixed package names at https://github.com/mroonga/mroonga/pull/348/commits/740eda20e97a1a43fac5e5e58ec0ed887059d424.

kou:
Let's try.

PR #352, opened 2020-09-15
doc developer: remove old contents

komainu8:
These contents are old.
We already don't use this development process.

May I remove these contents?


kou:
Yes.

komainu8:
I'll try again later.

PR #370, opened 2020-12-03
packages: add support for MariaDB 10.1 again only for Ubuntu

komainu8:
Because MariaDB 10.1 has supported yet on Ubuntu 18.04 LTS.


packages/mariadb-server-10.1-mroonga/Rakefile
```diff
@@ -0,0 +1,27 @@
...
+  def ubuntu_targets_default
+    [
+      ["bionic", "18.04"],
+    ]
+  end
+
+  def yum_targets_default
+    [
+    ]
+  end
```
kou, line 23:
```suggestion
  def enable_yum?
    false
  end
```

packages/mariadb-server-10.1-mroonga/Rakefile
```diff
@@ -0,0 +1,27 @@
...
+class MariaDBServer101MroongaPackageTask < MroongaPackageTask
+  def initialize
+    super("mariadb-server-10.1", "mariadb-10.1-mroonga")
+  end
+
+  def apt_targets_default
+    [
+      "debian-stretch",
+    ]
+  end
```
kou, line 12:
```suggestion
  def enable_apt?
    false
  end
```

Stretch reached EOL.

kou:
We can remove more Debian GNU/Linux Stretch related configurations and files.

PR #377, opened 2020-12-22
packages windows: Use an archive for uploading an artifact

komainu8:
Because if we don't use an archive for uploading an artifact, an
artifact uploading becomes slow.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -186,7 +186,6 @@ jobs:
       # Artifact
       - name: Compress the artifact
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -184,10 +184,16 @@ jobs:
             ..\mroonga\
 
       # Artifact
+      - name: Compress the artifact
+        run: |
+          ls mariadb-${{ matrix.mariadb-version }}-${{ matrix.package-platform }}
```
kou, line 189: Can we remove this?

komainu8: Thank you for your comment.
I've removed it.

PR #379, opened 2021-01-26
packages centos-7: use the old Apache Arrow

komainu8:
Because libgroonga 10.1.1 depends on libarrow 2.0.0.
However, currently the repository of Apache Arrow on CentOS7 provide
libarrow 3.0.0.

Therefore, we can't build Mroonga on CentOS7 now.
(e.g. https://github.com/mroonga/mroonga/runs/1767521129?check_suite_focus=true#step:12:1368)

This patch is temporary modification until release Groonga 11.0.0.


kou:
We'll use 11.0.0 not 10.2.0 for the next Groonga release.

komainu8:
Ah...
I made a mistake...

I understood. We'll use 10.2.0 for the next Groonga release.

kou:
We'll use 11.0.0. Because we release a new major version at every 02-09.

PR #380, opened 2021-01-27
packages mysql-server-8.0: use bundled zlib on centos7

komainu8:
1. mysql-server use the "crc32_z" function since version 8.0.23. This function is included zlib library, since version 1.2.9.
    If the version of zlib library is old that is provided on our OS (e.g. CentOS7), we can't build Mroonga.

2. mysql-server use the "z_size_t" type. This type is included "zconf.h".
    However, "zconf.h" does not exist in MySQL source directory.
    This header is generated in MySQL build directory after build MySQL.
    Therefore, we use the zlib library that is bundled MySQL 8.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -201,11 +201,9 @@ AC_DEFUN([CONFIG_OPTION_MYSQL],[
     mysql_protobuf_include_dir="$ac_mysql_source_dir/extra/protobuf/protobuf-3.11.4/src"
     MYSQL_INCLUDES="$MYSQL_INCLUDES -I$mysql_protobuf_include_dir"
   fi
-  if test -d "$ac_mysql_source_dir/extra/zlib"; then
+  if test -d "$ac_mysql_build_dir/extra/zlib"; then
     mysql_zlib_include_dir="$ac_mysql_source_dir/extra/zlib"
     MYSQL_INCLUDES="$MYSQL_INCLUDES -I$mysql_zlib_include_dir"
-  fi
-  if test -d "$ac_mysql_build_dir/extra/zlib"; then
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -201,6 +201,14 @@ AC_DEFUN([CONFIG_OPTION_MYSQL],[
...
     MYSQL_INCLUDES="$MYSQL_INCLUDES -I$mysql_protobuf_include_dir"
   fi
+  if test -d "$ac_mysql_source_dir/extra/zlib"; then
+    mysql_zlib_include_dir="$ac_mysql_source_dir/extra/zlib"
+    MYSQL_INCLUDES="$MYSQL_INCLUDES -I$mysql_zlib_include_dir"
+  fi
+  if test -d "$ac_mysql_build_dir/extra/zlib"; then
+    mysql_zlib_config_include_dir="$ac_mysql_build_dir/extra/zlib"
+    MYSQL_INCLUDES="$MYSQL_INCLUDES -I$mysql_zlib_config_include_dir"
+  fi
```
kou, line 211:
```suggestion
  if test -d "$ac_mysql_build_dir/extra/zlib"; then
    mysql_zlib_include_dir="$ac_mysql_source_dir/extra/zlib"
    MYSQL_INCLUDES="$MYSQL_INCLUDES -I$mysql_zlib_include_dir"
    mysql_zlib_config_include_dir="$ac_mysql_build_dir/extra/zlib"
    MYSQL_INCLUDES="$MYSQL_INCLUDES -I$mysql_zlib_config_include_dir"
  fi
```

kou:
`mysql_version_default` in `.spec.in` should be updated to use 8.0.23.

PR #384, opened 2021-02-04
packages mysql-server-5.6: drop mysql-server-5.6

komainu8:
Because it reached EOL at February 1, 2021


kou:
We also need to update https://mroonga.org/docs/install/centos.html .

komainu8:
Exactly.
I'll update it. Thank you for your comment.

PR #395, opened 2021-02-25
Don't use raw bitmap

komainu8:
Because the following patch changes function prototypes for "dbug_tmp_restore_column_maps" and "dbug_tmp_use_all_columns".

See: https://github.com/MariaDB/server/commit/21809f9a450df1bc44cef36377f96b516ac4a9ae

### DIFF
--- a/mrn_mysql_compat.h
+++ b/mrn_mysql_compat.h
@@ -1,7 +1,7 @@
 /* -*- c-basic-offset: 2 -*- */
 /*
   Copyright(C) 2011-2020 Sutou Kouhei <kou@clear-code.com>
-  Copyright(C) 2020 Horimoto Yasuhiro <horimoto@clear-code.com>
+  Copyright(C) 2020-2021 Horimoto Yasuhiro <horimoto@clear-code.com>
 
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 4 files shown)


mrn_mysql_compat.h
```diff
@@ -948,3 +948,10 @@ typedef HASH mrn_table_def_cache_type;
 #  define MRN_BITMAP_INIT(map, buf, n_bits) \
     bitmap_init((map), (buf), (n_bits), false)
 #endif
+
+#if defined(MRN_MARIADB_P) && (MYSQL_VERSION_ID >= 100237 || \
+                               MYSQL_VERSION_ID >= 100328 || \
+                               MYSQL_VERSION_ID >= 100418 || \
+                               MYSQL_VERSION_ID >= 100509)
+#  define DO_NOT_USE_RAW_BITMAP_FOR_DEBUG
```
kou, line 956: Could you add `MRN_` prefix for our macros?

mrn_mysql_compat.h
```diff
@@ -948,3 +948,10 @@ typedef HASH mrn_table_def_cache_type;
 #  define MRN_BITMAP_INIT(map, buf, n_bits) \
     bitmap_init((map), (buf), (n_bits), false)
 #endif
+
+#if defined(MRN_MARIADB_P) && (MYSQL_VERSION_ID >= 100237 || \
```
kou, line 952: This matches `100300` too.

mrn_mysql_compat.h
kou: How about `MRN_DBUG_TMP_USE_BITMAP_PP`?

kou:
`log opened` diff is unrelated of this.

PR #403, opened 2021-04-02
package yum: fix a name of test suite package for Percona Server 5.7

komainu8:
TODO: Test

### DIFF
--- a/packages/yum/test.sh
+++ b/packages/yum/test.sh
@@ -28,7 +28,7 @@ case ${package} in
     mroonga_package=$(echo "${package}" | sed -e '' -e 's/-server//g')
     service_name=mariadb
     ha_mroonga_so=ha_mroonga_official.so
-    package_prefix=MariaDB
+    test_package_name=MariaDB-test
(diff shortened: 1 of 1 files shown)


kou:
Can we do this in `case ${package} in ...`?

komainu8:
Is https://github.com/mroonga/mroonga/pull/403/commits/7b24d9809eb3a0cf578be721074a1b6b71ed201d as expected?

kou:
No. Here: https://github.com/mroonga/mroonga/blob/master/packages/yum/test.sh#L52-L68

packages/yum/test.sh
```diff
@@ -43,7 +43,7 @@ REPO
     mroonga_package=$(echo "${package}" | \
                         sed 's/-server-\([0-9]*\)\.\([0-9]*\)/\1\2-community/g')
     service_name=mysqld
-    package_prefix=mysql-community
+    testpackage_name=mysql-community-test
```
kou, line 46:
```suggestion
    test_package_name=mysql-community-test
```

PR #404, opened 2021-04-06
package: add support for downloading packages from GitHub release page 

komainu8:


packages/mroonga-package-task.rb
```diff
@@ -26,7 +19,6 @@ def initialize(mysql_package, rpm_package)
 
   def define
     super
-    define_windows_upload_task
   end
```
kou, line 22: Can we remove this method entirety?

packages/Rakefile
```diff
@@ -20,13 +20,13 @@ namespace :version do
   end
 end
 
-[:ubuntu, :windows].each do |target_namespace|
+[:ubuntu].each do |target_namespace|
   namespace target_namespace do
     desc "Upload #{target_namespace} packages"
     task :upload do
```
kou, line 26: Can we simplify task name to `ubuntu` from `ubuntu:upload`?

packages/Rakefile
```diff
@@ -39,7 +39,7 @@ end
     task :release do
```
kou, line 39: Can we simplify task name to `apt` from `apt:release`?

komainu8:
Thank you for your comments.
I've fixed them.

packages/Rakefile
```diff
@@ -20,28 +20,39 @@ namespace :version do
...
-        cd(package) do
-          ruby("-S", "rake", "#{target_namespace}:upload")
-        end
+namespace :ubuntu do
+  desc "Upload ubuntu packages"
+  task :upload do
+    packages.each do |package|
+      cd(package) do
+        ruby("-S", "rake", "ubuntu")
       end
```
kou, line 29:
```ruby
desc "Release Ubuntu packages"
task :ubuntu do
  packages.each do |package
    cd(package) do 
      ruby("-S", "rake", "ubuntu")
    end
  end
end
```

packages/Rakefile
kou: Ditto.

packages/Rakefile
```diff
@@ -20,39 +20,22 @@ namespace :version do
...
 end
 
-desc "Release ubuntu repositories"
-ubuntu_tasks = [
-  "ubuntu:upload",
-]
-task :ubunut => ubuntu_tasks
-
-
 [:apt, :yum].each do |target_namespace|
```
kou, line 32:
```suggestion
[:apt, :yum].each do |target|
```

Because we don't use it for namespace anymore.

packages/Rakefile
```diff
@@ -20,39 +20,22 @@ namespace :version do
...
-    task :release do
-      packages.each do |package|
-        cd(package) do
-          ruby("-S", "rake", "#{target_namespace}")
-        end
+  desc "Release #{target_namespace} packages"
+  task target_namespace do
+    packages.each do |package|
+      cd(package) do
+        ruby("-S", "rake", "#{target_namespace}")
```
kou, line 37:
```suggestion
        ruby("-S", "rake", target_namespace.to_s)
```

komainu8:
Thank you for your comments.
I've fixed them.

packages/mroonga-package-task.rb
```diff
@@ -169,71 +157,26 @@ def yum_expand_variable(key)
...
-            client.upload_asset(current_release.url, file_name, options)
-            rm(file_name)
-          end
-        end
+  def built_package_url(target_namespace, target)
+    url = "https://github.com/mroonga/mroonga/releases/download/v#{@version}/"
+    url << "packages-#{@mysql_package}-#{target}.zip"
+  end
+
+  def download_packages(target_namespace)
```
kou, line 169: If we use the same format as Groonga, we can reuse existing `download_packages` defined in `packages-groonga-org-package-task.rb`?

kou:
I've merged this change to master manually as 2e7eedae9007d1f651bba7bd30718284587aa0ac.
Because I couldn't push to komainu8:add-task-for-packages-download-from-github. (komainu8/mroonga has the same name tag.)

PR #405, opened 2021-04-14
package windows: add a patch for fixing syntax error in CMakeList

komainu8:
mariadb-10.2.36, 10.3.27, 10.4.17, 10.5.8 have a syntax error in
CMakeList of mariadb-connector-c.
(mariadb-connector-c is MariaDB client library.)

This bug has already fixed by
https://github.com/mariadb-corporation/mariadb-connector-c/commit/242cab8cbcd91af882233730a83627d3b12ced83
However, this fix is not included that above version of mariadb.

Therefore, we fix this bug by using this patch until next release for mariadb.
We will remove this patch when the next version of mariadb will be released.


packages/source/patches/mariadb-connector-c-fix-syntax-error-in-cmake-3.20.diff
```diff
@@ -0,0 +1,11 @@
...
++++ mariadb/libmariadb/cmake/ConnectorName.cmake	2021-04-14 16:52:43.049949830 +0900
+@@ -22,7 +22,7 @@
+     SET(MACHINE_NAME "x64")
+   ELSE()
+     SET(MACHINE_NAME "32")
+-  END()
++  ENDIF()
+ ENDIF()
+ 
+ SET(product_name "mysql-connector-c-${CPACK_PACKAGE_VERSION}-${PLATFORM_NAME}${CONCAT_SIGN}${MACHINE_NAME}")
```
kou, line 11: Could you use https://github.com/mariadb-corporation/mariadb-connector-c/commit/242cab8cbcd91af882233730a83627d3b12ced83.patch instead of creating a diff manually?

komainu8: Ah...I see.
I'll try it.

.github/workflows/windows.yml
```diff
@@ -102,6 +102,17 @@ jobs:
...
+          # (mariadb-connector-c is MariaDB client library.)
+          # This bug has already fixed by https://github.com/mariadb-corporation/mariadb-connector-c/commit/242cab8cbcd91af882233730a83627d3b12ced83
+          # However, this fix is not included that above version of mariadb.
+          # Therefore, we fix this bug by using this patch until next release for mariadb.
+          # We will remove this patch when the next version of mariadb will be released.
+          ridk exec curl `
+            -o mariadb-connector-c-fix-syntax-error-in-cmake-3.20.diff `
+            https://github.com/mariadb-corporation/mariadb-connector-c/commit/242cab8cbcd91af882233730a83627d3b12ced83.patch
+          ridk exec patch -d libmariadb -p1 -i `
+            ..\mariadb-connector-c-fix-syntax-error-in-cmake-3.20.diff
```
kou, line 115: Could you use the downloaded patch?

packages/source/patches/mariadb-connector-c-fix-syntax-error-in-cmake-3.20.diff
kou: Can we remove this file?

.github/workflows/windows.yml
```diff
@@ -102,6 +102,17 @@ jobs:
...
               ..\mroonga\packages\source\patches\mariadb-10.5.5-add-a-missing-export-symbol.diff
           }
+          # mariadb-10.2.36, 10.3.27, 10.4.17, 10.5.8 have a syntax error in CMakeList of mariadb-connector-c.
+          # (mariadb-connector-c is MariaDB client library.)
+          # This bug has already fixed by https://github.com/mariadb-corporation/mariadb-connector-c/commit/242cab8cbcd91af882233730a83627d3b12ced83
+          # However, this fix is not included that above version of mariadb.
+          # Therefore, we fix this bug by using this patch until next release for mariadb.
+          # We will remove this patch when the next version of mariadb will be released.
+          ridk exec curl `
+            -o mariadb-connector-c-fix-syntax-error-in-cmake-3.20.diff `
```
kou, line 112: `.patch` is better than `.diff` because the file file has more information such as author.

.github/workflows/windows.yml
komainu8: `mariadb-connector-c-fix-syntax-error-in-cmake-3.20.diff` that I'm using here is the downloaded patch.

It is renamed to `mariadb-connector-c-fix-syntax-error-in-cmake-3.20.diff` from `242cab8cbcd91af882233730a83627d3b12ced83.patch` .

Should I use the original patch file name (`242cab8cbcd91af882233730a83627d3b12ced83.patch`) ?
Because `mariadb-connector-c-fix-syntax-error-in-cmake-3.20.diff` is confusing.

.github/workflows/windows.yml
komainu8: I see.
I'll fix it.

.github/workflows/windows.yml
kou: Ah, `..\` is needed because we use `-d`.
I understand.

PR #407, opened 2021-04-28
configure: fix syntax

komainu8:


kou:
This is not a syntax error.
We need `;` only when we put `then` into the same line as condition: `if test "$with_debug" = "yes"; then`

komainu8:
I see.
Thank you for your comment.

PR #411, opened 2021-05-09
doc: use highlight directive instead of highlightlang directive

komainu8:
Because highlightlang directive is removed since Sphinx 4.0.0.


kou:
How about removing the directive entirely?
It seems that the `'default'` highlight language https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-highlight_language is enough for us. (Try `'python3'` first and then fallback to '`'none'`.)
Or we can set `highlight_lang = 'none'` in `doc/source/conf.py`.

PR #417, opened 2021-05-31
package ubuntu: package name change to ubuntu's package name

komainu8:
Mroonga's package name is changed  by the following commit.
https://github.com/mroonga/mroonga/commit/6bd846f2136fd7c657f232212cfc4759f9dc7e59

However, Ubuntu's MySQL package is mariadb-server-10.x and mysql-server-5.x.
Therefore, we need to convert Mroonga's package name to Ubuntu's package name.
Otherwise, we can't get version for MySQL and MariaDB for Ubuntu.

### DIFF
--- a/packages/mroonga-package-task.rb
+++ b/packages/mroonga-package-task.rb
@@ -89,6 +89,7 @@ class MroongaPackageTask < PackagesGroongaOrgPackageTask
     cache_key = [code_name, package_name]
     version = @ubuntu_package_versions[cache_key]
     return version if version
+    ubuntu_package_name = package_name.sub(/-/, "-server-")
     source_names = [code_name, "#{code_name}-updates"]
     source_names.each do |source_name|
       all_packages_url =
@@ -96,7 +97,7 @@ class MroongaPackageTask < PackagesGroongaOrgPackageTask
       URI.open(all_packages_url) do |all_packages|
         all_packages.each_line do |line|
           case line
-          when /\A#{Regexp.escape(package_name.sub(/-/, "-server-"))} \((.+?)[\s)]/o
+          when /\A#{Regexp.escape(ubuntu_package_name)} \((.+?)[\s)]/o
             version = $1
           end
         end
--- a/packages/mroonga-package-task.rb
+++ b/packages/mroonga-package-task.rb
@@ -84,17 +84,6 @@ class MroongaPackageTask < PackagesGroongaOrgPackageTask
     end
   end
 
-  def ubuntu_package_name(package_name)
-    ubuntu_package_names = {
-      "mariadb-10.2" => "mariadb-server-10.2",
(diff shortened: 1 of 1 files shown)


packages/mroonga-package-task.rb
```diff
@@ -84,6 +84,17 @@ def detect_mysql_version_debian(code_name)
...
   end
 
+  def ubuntu_package_name(package_name)
+    ubuntu_package_names = {
+      "mariadb-10.2" => "mariadb-server-10.2",
+      "mariadb-10.3" => "mariadb-server-10.3",
+      "mariadb-10.4" => "mariadb-server-10.4",
+      "mariadb-10.5" => "mariadb-server-10.5",
+      "mysql-5.7" => "mysql-server-5.7",
+    }
```
kou, line 94: It seems that we can generate Ubuntu package name from our package name: `package_name.sub(/-/, "-server")`

We don't want to maintain this mapping. If MariaDB or MySQL release a new series, we need to update this mapping.

komainu8: I see.
I'll fix it.

komainu8:
Thank you for your comment. I've fixed it.

packages/mroonga-package-task.rb
```diff
@@ -96,7 +96,7 @@ def detect_ubuntu_package_version(code_name, package_name)
       URI.open(all_packages_url) do |all_packages|
         all_packages.each_line do |line|
           case line
-          when /\A#{Regexp.escape(package_name)} \((.+?)[\s)]/o
+          when /\A#{Regexp.escape(package_name.sub(/-/, "-server-"))} \((.+?)[\s)]/o
```
kou, line 99: This will be difficult to understand in the future.

How about naming this value?

```ruby
ubuntu_package_name = package_name.sub(/-/, "-server-")
source_names = [...]
source_names.each do |...|
  ...
    ...
      when /\A#{Regexp.escape(ubuntu_package_name)} .../o
    ...
  ...
end
```

komainu8: Ah...
I see. I've fixed it at https://github.com/mroonga/mroonga/pull/417/commits/eda00db7225e03a0d3000fb7d56dab0a4717d933.

packages/mroonga-package-task.rb
```diff
@@ -89,14 +89,15 @@ def detect_ubuntu_package_version(code_name, package_name)
     cache_key = [code_name, package_name]
     version = @ubuntu_package_versions[cache_key]
     return version if version
+    ubuntu_package_name = package_name.sub(/-/, "-server-")
```
kou, line 92:
```suggestion
    # "mariadb-10.5" -> "mariadb-server-10.5"
    # "mysql-5.7" -> "mysql-server-5.7"
    ubuntu_package_name = package_name.sub(/-/, "-server-")
```

PR #424, opened 2021-06-21
package mariadb-10.3: rebuild against the latest MariaDB 10.3 package

komainu8:
We need to increase minor version of Debian package for Mroonga when we update Debian package for Mroonga.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -138,12 +138,21 @@ jobs:
           restore-keys: package-${{ matrix.os }}-${{ matrix.package }}-ccache-
       - name: Build with Docker
         run: |
+          case ${{ matrix.os }} in
+            debian-*)
+              deb_release=$(head -n1 packages/${{ matrix.package }}-mroonga/debian/changelog | \
+                            awk '{print $2}' | \
+                            awk -F "-" '{print $2}' | \
+                            sed -e 's/[^0-9]//g')
+              ;;
+          esac
           cd packages/${{ matrix.package }}-mroonga
           rake ${{ matrix.package-type }}:build BUILD_DIR=build
         env:
           APT_TARGETS: ${{ matrix.os }}
           GROONGA_REPOSITORY: ../../../groonga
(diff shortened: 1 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -138,8 +138,16 @@ jobs:
...
+            debian-*)
+              deb_release=$(head -n1 packages/${{ matrix.package }}-mroonga/debian/changelog | \
+                            awk '{print $2}' | \
+                            awk -F "-" '{print $2}' | \
+                            sed -e 's/[^0-9]//g')
+              ;;
+          esac
           cd packages/${{ matrix.package }}-mroonga
-          rake ${{ matrix.package-type }}:build BUILD_DIR=build
+          rake ${{ matrix.package-type }}:build BUILD_DIR=build DEB_RELEASE=${deb_release}
```
kou, line 150: Why do we need to specify `DEB_RELEASE` here?
`DEB_RELEASE` is for `rake version:update` not `rake apt:build`.

kou:
CI is failed because of the target branch isn't master.
This will work on master and `*-maintainance` branch.

.github/workflows/linux.yml
```diff
@@ -119,6 +119,8 @@ jobs:
         run: |
           make dist
       - name: Update version
+        if: |
+          github.ref == 'refs/heads/master'
```
kou, line 123: Renaming maintenance branch name rule to `maintenance/11.03` from `v11.03-maintenance` and changing this to `startsWith(github.ref, 'refs/tags/') || startsWith(github.ref, 'refs/heads/maintenance/')` may be better.

komainu8: "Update version" execute only on master.
So, I think that the following code is correct. Is that my misunderstanding?

```
if: |
  !startsWith(github.ref, 'refs/tags/') && !startsWith(github.ref, 'refs/heads/maintenance/')
```

kou: Right.
My code was wrong.

PR #425, opened 2021-06-21
ci: add support for "maintenance/XX.XX" branch name

komainu8:
Because we will rename maintenance branch name to "maintenance/XX.XX" from "vXX.XX-maintenance".

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,13 +1,8 @@
 name: Linux
 on:
   push:
-    branches:
-      - "**"
-    tags:
-      - "*"
   pull_request:
     branches:
-      - "*"
   schedule:
     - cron: |
         0 0 * * *
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -2,7 +2,7 @@ name: Linux
 on:
   push:
     branches:
-      - "*"
+      - "**"
     tags:
       - "*"
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/linux.yml
```diff
@@ -2,7 +2,7 @@ name: Linux
 on:
   push:
     branches:
-      - "*"
+      - "**"
     tags:
       - "*"
```
kou, line 7:
```yaml
branches:
tags:
```

will work.

komainu8: Oh...I see.

"on.push" event only has branches parameter and tags parameter.
the following definition is the same as "on: push:"

```
on:
  push:
    branches:
    tags:
```

Therefore, I've fixed as below.

```
on:
  push:
```

The following definition is also the same as "pull_request: branches:". 　I also fixed this.

.github/workflows/linux.yml
```diff
@@ -120,7 +115,8 @@ jobs:
           make dist
       - name: Update version
         if: |
-          github.ref == 'refs/heads/master'
+          !startsWith(github.ref, 'refs/tags/') &&
+          !startsWith(github.ref, 'refs/heads/maintenance/')
```
kou, line 119: Or we keep using `vXX.YY-maintenance`:

```suggestion
          !endsWith(github.ref, '-maintenance')
```

komainu8: You mean we need this modification if we keep using `vXX.YY-maintenance` branch?
If we use `maintenace/XX.YY` branch, is this modification needless?

I will use `maintenace/XX.YY` branch.

kou: OK.

PR #445, opened 2021-09-21
ci: enable test for Mroonga for MariaDB 10.5 on Debian 11

komainu8:
TODO:Test


kou:
Enabled but there is a problem:

https://github.com/mroonga/mroonga/runs/4185052001?check_suite_focus=true#step:5:1631

> +Warnings:
> +Warning	1620	Plugin is busy and will be uninstalled on shutdown

PR #447, opened 2021-09-27
test: use Mroonga on storage engine

komainu8:
This test is for confirming whether fixed a bug of [groonga-dev,04855]
or not. However, this bug doesn't reproduce if we don't use Mroonga on
storage engine.

### DIFF
--- a/mysql-test/mroonga/storage/index/primary/varchar/t/in_select_empty_string.test
+++ b/mysql-test/mroonga/storage/index/primary/varchar/t/in_select_empty_string.test
@@ -25,7 +25,7 @@ DROP TABLE IF EXISTS ids;
 CREATE TABLE ids (
   id varchar(7) PRIMARY KEY,
   parent_id varchar(7)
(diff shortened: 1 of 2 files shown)


mysql-test/mroonga/storage/index/primary/varchar/t/in_select_empty_string.test
```diff
@@ -25,7 +25,7 @@ DROP TABLE IF EXISTS ids;
 CREATE TABLE ids (
   id varchar(7) PRIMARY KEY,
   parent_id varchar(7)
-) DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
+) ENGINE=Mroonga DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```
kou, line 28: This is needless. Because we use Mroonga as the default storage engine in test.

komainu8: I see.
Thank you for your comment.

PR #460, opened 2021-11-16
ci: add a test job for maintenance branch

komainu8:
TODO: Probably I can be simpler it.

The maintenance branch makes the same version packages.
Therefore, an upgrade test always fails.

We execute the upgrade test on only master.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -292,6 +292,8 @@ jobs:
         run: |
           vagrant up ${{ matrix.os }}
       - name: Run test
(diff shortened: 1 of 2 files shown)


kou:
The tests should not be failed because the maintenance branch packages must be newer (`11.09-2`) than the released packages (`11.09-1`).
We should increment `Release` in the maintenance branch.

komainu8:
This PR is needless.
Because I increment `Release` in the maintenance branch.

 Therefore, I close this PR

PR #464, opened 2021-11-22
ci: enable almalinux 8

komainu8:
Because groonga-normalizer-mysql for AlmaLinux 8 has been already released.


kou:
We can merge this once CI is green.

PR #470, opened 2021-12-01
Fix a typo

marclaporte:
I don't know if some other string must be corrected to avoid breaking the translation.


kou:
We can detect it automatically by a gettext feature.

PR #475, opened 2021-12-22
ci install: enable Ubuntu 21.10 (Impish)

komainu8:
Because Ubuntu Impish has been already available on bento.


komainu8:
Tests of this modification passed at https://github.com/komainu8/mroonga/actions/runs/1609923376.

PR #478, opened 2022-01-21
packages mysql-community: use the latest release

komainu8:
This is testing now.

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -12024,6 +12024,10 @@ void ha_mroonga::storage_store_field_timestamp(Field *field,
   struct timeval time_value;
   GRN_TIME_UNPACK(time, time_value.tv_sec, time_value.tv_usec);
   timestamp_field->store_timestamp(&time_value);
+#elif defined(MRN_TIMESTAMP_USE_MY_TIMEVAL)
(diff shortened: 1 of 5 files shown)


komainu8:
@kou 
I have a question.
Are https://github.com/mroonga/mroonga/blob/master/lib/mrn_timestamp_field_value_converter.hpp#L43-L47 correct?

Because if `field_->get_timestamp(&time_value, &warnings)` is `true`, Groonga is not set value to `grn_time` that is retuen value. Therfore, the return value is `0` even if `field_->get_timestamp(&time_value, &warnings)` is `true`.

In addition, if field_->get_timestamp(&time_value, &warnings) is false, the return value is also 0, probably.

I think that the return value is `0`in any case.

kou:
`Field_timestamp::get_timestamp()` returns "`false`" not "`true`" if it sets value successfully.

komainu8:
> `Field_timestamp::get_timestamp()` returns "`false`" not "`true`" if it sets value successfully.

Oh...I see. Thank you for your comment.

komainu8:
The tests of this PR are passed as below.
https://github.com/mroonga/mroonga/actions/runs/1755111653.
https://github.com/mroonga/mroonga/actions/runs/1755111652.

PR #489, opened 2022-02-14
ci: enable upgrade test for AlmaLinux

komainu8:
I'll merge if the CI has passed.

### DIFF
--- a/packages/yum/test.sh
+++ b/packages/yum/test.sh
@@ -177,11 +177,11 @@ sudo ${DNF} erase -y \
   "${mysql_package_prefix}-*"
 
 # Disable upgrade test for first time packages.
-case ${os}-${package} in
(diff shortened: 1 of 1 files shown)


packages/yum/test.sh
```diff
@@ -177,11 +177,11 @@ sudo ${DNF} erase -y \
...
-case ${os}-${package} in
-  almalinux-*) # TODO: Remove this after 11.10 release.
-    exit
-    ;;
-esac
+#case ${os}-${package} in
+#  almalinux-*) # TODO: Remove this after 11.10 release.
+#    exit
+#    ;;
+#esac
```
kou, line 184: We should remove this comment block entirely.

komainu8:
This modification has already pushed master.
Therefore, this is needless. I close this PR.

PR #495, opened 2022-04-04
ci linux: enable "Test" for Ubuntu Jammy

komainu8:
Because Ubuntu Jammy has already available on bento.

### DIFF
--- a/packages/apt/test.sh
+++ b/packages/apt/test.sh
@@ -127,25 +127,39 @@ for test_suite_name in $(find plugin/mroonga -type d '!' -name '[tr]'); do
 done
 set -x
 
-sudo \
-  ./mtr \
-  --force \
-  --no-check-testcases \
-  --parallel=${parallel} \
-  --retry=3 \
-  --suite="${test_suite_names}"
-
-case ${package} in
-  mariadb-*)
-    # Test with binary protocol
+case ${distribution} in
+  ubuntu)
     sudo \
       ./mtr \
       --force \
       --no-check-testcases \
       --parallel=${parallel} \
-      --ps-protocol \
+      --retry=3 \
+      --client-bindir="${mysql_test_dir}/bin" \
+      --suite="${test_suite_names}"
+    ;;
+  *)
+    sudo \
+      ./mtr \
+      --force \
+      --no-check-testcases \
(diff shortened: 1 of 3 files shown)


packages/apt/test.sh
```diff
@@ -12,11 +12,20 @@ echo "debconf debconf/frontend select Noninteractive" | \
...
+  Ubuntu)
+    repository=universe
+  ;;
+esac
 code_name=$(lsb_release --codename --short)
 architecture=$(dpkg --print-architecture)
 
 wget \
-  https://packages.groonga.org/debian/groonga-apt-source-latest-${code_name}.deb
+  https://packages.groonga.org/${distribution,}/groonga-apt-source-latest-${code_name}.deb
```
kou, line 28:
```suggestion
  https://packages.groonga.org/${distribution}/groonga-apt-source-latest-${code_name}.deb
```

packages/apt/test.sh
```diff
@@ -66,7 +75,7 @@ Architectures: amd64 source
 SignWith: ${GPG_KEY_ID}
 DISTRIBUTIONS
 reprepro includedeb ${code_name} \
-  ${repositories_dir}/debian/pool/${code_name}/main/*/*/*_{${architecture},all}.deb
+  ${repositories_dir}/${distribution,}/pool/${code_name}/${repository}/*/*/*_{${architecture},all}.deb
```
kou, line 78:
```suggestion
  ${repositories_dir}/${distribution}/pool/${code_name}/${repository}/*/*/*_{${architecture},all}.deb
```

packages/apt/test.sh
komainu8: `lsb_release --id --short` returns a string that is the upper case for only the first character (e.g. `Ubuntu`, `Debian`.)
Therefore, we can't use it in the URL as it is now. 
I change `$(lsb_release --id --short)` to the lower case with `{distribution,}`.

kou: Oh, I didn't know `${XXX,}` feature!
But it's a bit tricky... How about converting `lsb_release --id --short` output?

```bash
distribution=$(lsb_release --id --short | tr 'A-Z' 'a-z')
```

PR #501, opened 2022-05-06
ci linux build: add a task for building packages for Ubuntu

komainu8:


kou:
It seems that MySQL 8.0 doesn't provide packages for Ubuntu 22.04 yet: https://repo.mysql.com/apt/ubuntu/dists

BTW, is it intentional that you do it for `mysql-community-8.0-mroonga` not `mysql-8.0-mroonga`?

komainu8:
> It seems that MySQL 8.0 doesn't provide packages for Ubuntu 22.04 yet: https://repo.mysql.com/apt/ubuntu/dists

Oh...I see. Thank you for your comment.
I disable Ubuntu 22.04 temporaliy.

> BTW, is it intentional that you do this for mysql-community-8.0-mroonga not mysql-8.0-mroonga?

Ah...Yes... This is intentional.
However, I should have implemented it on mysql-8.0-mroonga at first...
I think that I want to support mysql-community-8.0-mroonga and mysql-8.0-mroonga  as with Debian.

PR #507, opened 2022-06-14
packages mariadb-10.2: drop

komainu8:
Because MariaDB 10.2 reached EOL.


komainu8:
All CI passed as below.

https://github.com/mroonga/mroonga/actions/runs/2493297404
https://github.com/mroonga/mroonga/actions/runs/2493297409

Therefore, I marge this PR.

PR #508, opened 2022-06-14
ci linux: add "Test" for Ubuntu Jammy

komainu8:
This is testing now.


kou:
Close in favor of #495.

PR #513, opened 2022-07-01
packages: add tests whether Mroonga is registered correctly

komainu8:
Because CI passed even if the SQL to install Mroonga is fail.

### DIFF
--- a/packages/yum/test.sh
+++ b/packages/yum/test.sh
@@ -111,8 +111,6 @@ esac
 
 # Install
 function mroonga_is_registered() {
-  have_auto_generated_password=$1
-
   sudo systemctl start ${service_name}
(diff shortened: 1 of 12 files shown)


packages/yum/test.sh
```diff
@@ -216,3 +236,8 @@ esac
 sudo ${DNF} install -y ${old_package}
 sudo ${DNF} install -y \
   ${repositories_dir}/${os}/${major_version}/*/Packages/*.rpm
+
+mroonga_exist ${have_auto_generated_password}
+if [ $? -ne 0 ]; then
+  exit 1
+fi
```
kou, line 243: Why do we need this? We already have `set -e`.

Issue #509, opened 2022-06-14
Outdated instructions to install from source


perrinjerome:
In https://mroonga.org/docs/install/others.html#mysql-u-root-install-sql we can read:

> You need to run some SQLs to register Mroonga to MySQL such as `INSTALL PLUGIN` and `CREATE FUNCTION`. They are written in `${PREFIX}/share/mroonga/install.sql`.
>
> Here is an example when you specify `--prefix=$HOME/local` to `configure`:
>
>```
>% mysql -u root < $HOME/local/share/mroonga/install.sql
>```

but since 12.04 ( https://github.com/mroonga/mroonga/commit/35f593037c2d8f5a497349ff98f3a3ce31ee95de ) it is also required to run `update.sql`.

If it's possible to keep the current behavior of using only `install.sql`, I'm sure people packaging mroonga will appreciate ;)

Thanks

komainu8:
@perrinjerome 
Thank you for your report.

I'm fixing that we don't require `update.sql` when the update and manual install for Mroonga at https://github.com/mroonga/mroonga/pull/511.

perrinjerome:
@komainu8 @kou I wanted to confirm that with the just released mroonga 12.0.7 everything is fine regarding this and to thank you one more time 👍

kou:
That's good to know! :-)

PR #522, opened 2022-08-15
test: enable upgrade test for Debian bullseye and Ubuntu jammy

komainu8:
GitHub: fix https://github.com/mroonga/mroonga/issues/520, https://github.com/mroonga/mroonga/issues/525

I become use `install.sql` when Mroonga installs or upgrades by this modification. 
(We use `install_plugin.sql` and `update.sql` instead of `install.sql` when Mroonga installs and upgrades until now...
Therefore, we couldn't detect problems #520 and #525.)

We can also detect problems with Mroonga's install script in CI by this modification.


komainu8:
@kou Could you review this PR?

configure.ac
```diff
@@ -32,6 +32,9 @@ AC_SUBST([MRN_PACKAGE_STRING])
 MRN_BUNDLED=FALSE
 AC_SUBST([MRN_BUNDLED])
 
+MRN_DATA_DIR="/usr/share/mroonga"
```
kou, line 35:
```suggestion
MRN_DATA_DIR="${datadir}/${PACKAGE}"
```

kou: Could you also define `MRN_DATA_DIR` in `CMakeLists.txt`?

data/deb/postinst.sh
```diff
@@ -70,7 +70,7 @@ install_mroonga() {
     need_manual_update=yes
   fi
 
-  install_plugin_sql=/usr/share/mroonga/install_plugin.sql
```
kou, line 73: Can we remove `data/install_plugin.sql.in`?

packages/apt/test.sh
```diff
@@ -165,11 +165,20 @@ esac
 
 # Upgrade
 if [ -n "${old_package}" ]; then
+  cd
```
kou, line 168: Could you restore the current working directory after we run test?

```diff
diff --git a/packages/apt/test.sh b/packages/apt/test.sh
index f9162b2d..57b35d2a 100755
--- a/packages/apt/test.sh
+++ b/packages/apt/test.sh
@@ -107,7 +107,7 @@ sudo apt install -V -y \
   ${test_package} \
   patch
 
-cd ${mysql_test_dir}
+pushd ${mysql_test_dir}
 sudo rm -rf plugin/mroonga
 sudo mkdir -p plugin
 sudo cp -a /vagrant/mysql-test/mroonga/ plugin/
@@ -162,6 +162,7 @@ case ${package} in
     sudo ./mtr "${mtr_args[@]}" --ps-protocol
     ;;
 esac
+popd
 
 # Upgrade
 if [ -n "${old_package}" ]; then
```

packages/apt/test.sh
```diff
@@ -165,11 +165,20 @@ esac
...
     ${package} \
     "${mysql_package_prefix}-*"
   sudo rm -rf /var/lib/mysql
   sudo mv /etc/apt/sources.list.d/${package}.list /tmp/
+  case ${package} in
+    mysql-community-8.*)
+      sudo \
+        env DEBIAN_FRONTEND=noninteractive \
+            MYSQL_SERVER_VERSION=mysql-${mysql_version} \
+          apt install -y ./mysql-apt-config_*_all.deb
```
kou, line 179: Why is `mysql-apt-config*.deb` removed? Is `"${mysql_package_prefix}-*" matched?

kou: `${mysql_package_prefix}` is `mysql` and it matches `mysql-apt-config`:

https://github.com/mroonga/mroonga/runs/7985783662?check_suite_focus=true#step:5:2377

```text
+ sudo apt purge -V -y mysql-community-8.0-mroonga 'mysql-*'
```

https://github.com/mroonga/mroonga/runs/7985783662?check_suite_focus=true#step:5:2476

```text
   mysql-apt-config* (0.8.22-1)
```

komainu8: Yes, `mysql-apt-config*.deb` matched "${mysql_package_prefix}-*".
Therefore, `mysql-apt-config` is removed by `sudo apt purge -V -y mysql-community-8.0-mroonga 'mysql-*'`.

data/deb/postinst.sh
komainu8: Why do we can remove data/install_plugin.sql.in?
install.sql call install_plugin.sql. Therefore, I think we need data/install_plugin.sql.in.

configure.ac
```diff
@@ -32,6 +32,9 @@ AC_SUBST([MRN_PACKAGE_STRING])
 MRN_BUNDLED=FALSE
 AC_SUBST([MRN_BUNDLED])
 
+MRN_DATA_DIR="${datadir}/${PACKAGE}"
+AC_SUBST([MRN_DATA_DIR])
```
perrinjerome, line 36: @komainu8  In the github actions from this build ( https://github.com/mroonga/mroonga/runs/8012897950?check_suite_focus=true#step:9:165 ), we can see this warning:

```
config.status: creating data/install.sql
config.status: WARNING: data/install.sql contains a reference to the variable `datarootdir'
which seems to be undefined.  Please make sure it is defined
```

I believe the resulting `data/ install.sql` must contain something like this

```
source ${datarootdir}/mroonga/install_plugin.sql
source ${datarootdir}/mroonga/update.sql
```

isn't it ?

Actually, before I noticed that there was already this work, I tried to update to mroonga 12.06 and noticed that the fix for https://github.com/mroonga/mroonga/issues/509 was not functional when using `configure` and `make`. Basically, I observed what was also reported in https://github.com/mroonga/mroonga/issues/520 .

So I tried a bit myself. I started with the same change as this one, using `AC_SUBST` in configure and noticed that problem, so I studied a bit more autoconf / automake - but I don't have much experience with these tools and they are extremely complex, so before you read I have to warn that I might be wrong 😉 . Here's what I found:

According to https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/Installation-Directory-Variables.html , we can not simply use `AC_SUBST` to let configure substitute `MRN_DATA_DIR`, because it is an installation directory variable and the substitution needs to be done with make ( this is the [*you should not use these variables except in makefiles*](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/Installation-Directory-Variables.html#:~:text=you%20should%20not%20use%20these%20variables%20except%20in%20makefiles) part). It seems it is so that a different directory can be selected during make install. According to my understanding of this documentation, the substitution should not happen during the `configure` step, it should happen during the `make` step. 

I came up with this patch https://github.com/perrinjerome/mroonga/commit/525c4a16c6a11b2474866806ebfffaedbde5ef49  and then I found out that you were already working on this, so I preferred to discuss instead of opening another pull request for the same problem. If you find it helpful to integrate this patch (or part of this patch) please go ahead, if you like I can also make a pull request with this patch. But I am a bit surprised that this was so complicated, so maybe I was doing something wrong or maybe it's just that autoconf/automake is very complex.

In any case, thank you for working on this issue.

data/deb/postinst.sh
kou: If this was the last place that refers `install_plugin.sql` except `install.sql`, we don't need to split `install_plugin.sql` from `install.sql`.
We can merge `install_plugin.sql` into `install.sql`.

CMakeLists.txt
```diff
@@ -529,7 +529,7 @@ configure_file(
 if(MRN_BUNDLED)
   set(MRN_DATA_DIR "${INSTALL_MYSQLSHAREDIR}/${PROJECT_NAME}")
 else()
-  set(MRN_DATA_DIR "share/${PROJECT_NAME}")
+  set(MRN_DATA_DIR "/usr/share/${PROJECT_NAME}")
```
kou, line 532: We should not hard-code the path.
This doesn't work with `-DCMAKE_INSTALL_PREFIX=/usr/local`.

configure.ac
kou: Thanks.
We will merge your change to this pull request.

PR #539, opened 2022-09-07
udf highlight_html: add support specify tag

komainu8:
GitHub: fix GH-537

We can specify `open_tag` and `close_tag` for highlighted keywords with this change.

Syntax:

    mroonga_highlight_html(target_text,
                           keyword0, keyword1, ...,
                           option0, option1, ...)

    target_text: A string that has 0 or more keywords to be highlighted

    keywordN: A string to be highlighted
      * If "query" option is specified,
        you don't need to specify any keywords
      * If "query" option isn't specified,
        you need to specify at least one keyword.

    optionN: A string to be highlighted
      * Syntax: "value AS name"
      * Available names:
        * open_tag: A string to be used as an open tag
          (default: '<span class="keyword">')
        * close_tag: A string to be used as a close tag
          (default: '</span>')
        * query: A string to be used as a query that has keywords.
          Highlighted keywords are extracted from the query.
          (default: NULL)

Samples:

    -- 2 keywords: Mroonga and Groonga
    -- No option
    SELECT mroonga_highlight_html(
             'Mroonga is the Groonga based storage engine.',
             'Mroonga', 'Groonga') AS highlighted;
    -- highlighted
    -- <span class="keyword">Mroonga</span> is the
    -- <span class="keyword">Groonga</span> based storage engine.

    -- 2 keywords: Mroonga and Groonga
    -- 2 options: open_tag (<mark>) and close_tag (</mark>)
    SELECT mroonga_highlight_html(
             'Mroonga is the Groonga based storage engine.',
             'Mroonga', 'Groonga',
             '<mark>' AS open_tag,
             '</mark>' AS close_tag) AS highlighted;
    -- highlighted
    -- <mark>Mroonga</mark> is the <mark>Groonga</mark> based storage engine.

    -- No keyword: (Extracted from query: Mroonga and Groonga)
    -- 3 options: open_tag (<mark>),
    --            close_tag (</mark>) and
    --            query (Mroonga OR Groonga)
    SELECT mroonga_highlight_html(
             'Mroonga is the Groonga based storage engine.',
             '<mark>' AS open_tag,
             '</mark>' AS close_tag,
             'Mroonga OR Groonga' AS query) AS highlighted;
    -- highlighted
    -- <mark>Mroonga</mark> is the <mark>Groonga</mark> based storage engine.

We can change the default open tag (`<span class="keyword">`) and
close tag (`</span>`) by `open_tag` option and `close_tag`.
In general, you should specify both of `open_tag` option and `close_tag`
option but you can specify one of them.


udf/mrn_udf_highlight_html.cpp
```diff
@@ -295,6 +301,21 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
     info->query_mode.default_column = NULL;
   }
 
+  info->specify_tag.used = false;
+  if (args->arg_count == 4) {
```
HashidaTKS, line 305: Does it work if multiple keywords specified like below?

```
SELECT mroonga_highlight_html('Mroonga is the Groonga based storage engine.',
                              'groonga', 'mroonga',
                              '<mark>' AS open_tag, '</mark>' AS close_tag) AS highlighted;
```

HashidaTKS: Ah, I found the test for the multiple keywords.
But why they work as expected...?

komainu8: Thank you for your comment.

You are right. Currently, this function does not work if multiple keywords are specified.
Tests of this function are not executed now. Because this modification failed to build yet.

HashidaTKS: I see. Thenk you for your reply!

HashidaTKS:
We can fix failing of `Windows / Package` checks by adding `#include <string>` into `mrn_udf_highlight_html.cpp`.
`std::string` recuires `string` on Windows.

I have fixed the same issue on https://github.com/mroonga/mroonga/pull/540/commits/1b8583d1f8f607ab5b90d164c6a2c6155bdcf581 : ) .

HashidaTKS:
@komainu8 @kou 

Would you review?

HashidaTKS:
I noticed that this don't work correctly when we specify `AS query` with `open_tag` and `close_tag`.
And also maybe open_tag and close_tag is added to keywords...
I will fix them.

HashidaTKS:
@kou @komainu8 

Would you review this when you have time?

Some tests failed but I think those tests are not related to this change.

udf/mrn_udf_highlight_html.cpp
```diff
@@ -46,13 +47,20 @@ typedef struct st_mrn_highlight_html_info
     bool used;
     grn_obj *table;
     grn_obj *default_column;
+    const char* query;
+    unsigned long query_length;
```
kou, line 51: It seems that we don't need to add them.

udf/mrn_udf_highlight_html.cpp
```diff
@@ -46,13 +47,20 @@ typedef struct st_mrn_highlight_html_info
...
   } query_mode;
+  struct {
+    bool used;
+    const char *open_tag;
+    const char *close_tag;
+  } specify_tag;
```
kou, line 57: How about defining them in top-level and use them for the default tags (`<span class="keyword">` and `</span>`) too?

```suggestion
  const char *open_tag;
  const char *close_tag;
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -286,13 +304,40 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
...
+    if (args->attribute_lengths[i] == strlen("query") &&
+        strncmp(args->attributes[i], "query", strlen("query")) == 0) {
+      info->query_mode.used = true;
+      info->query_mode.table = NULL;
+      info->query_mode.default_column = NULL;
+      info->query_mode.query = args->args[i];
+      info->query_mode.query_length = args->lengths[i];
+    }
+    if (attribute_length == strlen("open_tag") &&
+        strncmp(attribute, "open_tag", strlen("open_tag")) == 0) {
```
kou, line 321:
```suggestion
    } else if (attribute_length == strlen("open_tag") &&
               strncmp(attribute, "open_tag", strlen("open_tag")) == 0) {
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -286,13 +304,40 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
...
+  if (info->specify_tag.used) {
+    if (info->specify_tag.open_tag == nullptr) {
+      snprintf(message, MYSQL_ERRMSG_SIZE,
+       "mroonga_highlight_html(): missing the required \"open_tag\" argument.");
+      goto error;
+    } else if (info->specify_tag.close_tag == nullptr) {
+      snprintf(message, MYSQL_ERRMSG_SIZE,
+       "mroonga_highlight_html(): missing the required \"close_tag\" argument.");
+      goto error;
+    }
```
kou, line 340: How about using `<span class="keyword">` and `</span>` as the default?

mysql-test/mroonga/storage/function/highlight_html/t/open_tag_close_tag.test
```diff
@@ -0,0 +1,26 @@
...
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+--source ../../../../include/mroonga/not_embedded.inc
+--source ../../../../include/mroonga/have_mroonga.inc
+--source ../../../../include/mroonga/load_mroonga_functions.inc
+
+SELECT mroonga_highlight_html('Mroonga is the Groonga based storage engine.',
+                              'groonga',
+                              '<mark>' AS open_tag, '</mark>' AS close_tag) AS highlighted;
```
kou, line 23: How about using a valid HTML tag such as `<b>` and `</b>`?

kou: Could you also add the same text as `open_tag` and `close_tag` in the target text to confirm that `open_tag` and `close_tag` aren't used as keyword?

HashidaTKS: `<mark>` seems to be a valid HTML tag.

https://www.w3schools.com/TAGS/tag_mark.asp

HashidaTKS: > Could you also add the same text as `open_tag` and `close_tag` in the target text to confirm that `open_tag` and `close_tag` aren't used as keyword?

The test for that cases was already added.
Please confirm it.

https://github.com/mroonga/mroonga/blob/8dd904870fe721804dde8dca4ce719f9d8edcf49/mysql-test/mroonga/storage/function/highlight_html/t/open_tag_close_tag_not_added_to_keywords.test

[mysql-test/mroonga/storage/function/highlight_html/r/open_tag_close_tag_missing_close_tag.result](https://github.com/mroonga/mroonga/blob/8dd904870fe721804dde8dca4ce719f9d8edcf49/mysql-test/mroonga/storage/function/highlight_html/r/open_tag_close_tag_missing_close_tag.result)

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

I have changed the specification of `AS query` like below as I have written in the description of this PR.

> From this change, AS query also can be specified at anywhere after the second argument.

So we should use them instead of `args->args[1]` and `args->lengths[1]` when  `AS query` is specified like the following commit.

https://github.com/mroonga/mroonga/pull/539/commits/663293359d9891918aee252175a919eb0369909d

mysql-test/mroonga/storage/function/highlight_html/t/open_tag_close_tag.test
kou: > `<mark>` seems to be a valid HTML tag.

Wow! I didn't know that!
We should have used it for the default tag...

kou: > > Could you also add the same text as `open_tag` and `close_tag` in the target text to confirm that `open_tag` and `close_tag` aren't used as keyword?
> 
> The test for that case was already added. Please confirm it.
> 
> https://github.com/mroonga/mroonga/blob/8dd904870fe721804dde8dca4ce719f9d8edcf49/mysql-test/mroonga/storage/function/highlight_html/t/open_tag_close_tag_not_added_to_keywords.test

Could you use different text such as `[` not `open_tag` to ensure `open_tag` argument is used?

```sql
SELECT mroonga_highlight_html('Groonga should be highlighted but [ and ] should not be highlighted',
                              'Groonga', '[' AS open_tag, ']' AS close_tag) AS highlighted;
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -168,6 +173,16 @@ static mrn_bool mrn_highlight_html_prepare(mrn_highlight_html_info *info,
       if (!args->args[i]) {
         continue;
       }
+      const char *attribute = args->attributes[i];
+      unsigned long attribute_length = args->attribute_lengths[i];
```
kou, line 177:
```suggestion
      auto attribute = args->attributes[i];
      auto attribute_length = args->attribute_lengths[i];
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -287,12 +302,28 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
...
 
-  if (args->arg_count == 2 &&
-      args->attribute_lengths[1] == strlen("query") &&
-      strncmp(args->attributes[1], "query", strlen("query")) == 0) {
-    info->query_mode.used = true;
-    info->query_mode.table = NULL;
-    info->query_mode.default_column = NULL;
+  for (unsigned int i = 0; i < args->arg_count; i++) {
+    const char *attribute = args->attributes[i];
+    unsigned long attribute_length = args->attribute_lengths[i];
```
kou, line 307:
```suggestion
    auto attribute = args->attributes[i];
    auto attribute_length = args->attribute_lengths[i];
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -287,12 +302,28 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
 
   info->query_mode.used = false;
 
-  if (args->arg_count == 2 &&
-      args->attribute_lengths[1] == strlen("query") &&
-      strncmp(args->attributes[1], "query", strlen("query")) == 0) {
-    info->query_mode.used = true;
-    info->query_mode.table = NULL;
-    info->query_mode.default_column = NULL;
+  for (unsigned int i = 0; i < args->arg_count; i++) {
```
kou, line 305:
```suggestion
  for (unsigned int i = 1; i < args->arg_count; i++) {
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -287,12 +302,28 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
...
+    if (args->attribute_lengths[i] == strlen("query") &&
+        strncmp(args->attributes[i], "query", strlen("query")) == 0) {
+      info->query_mode.used = true;
+      info->query_mode.table = NULL;
+      info->query_mode.default_column = NULL;
+      info->query_mode.query = args->args[i];
+      info->query_mode.query_length = args->lengths[i];
+    } else if (attribute_length == strlen("open_tag") &&
+               strncmp(attribute, "open_tag", strlen("open_tag")) == 0) {
+      info->open_tag = args->args[i];
```
kou, line 317: This is unsafe.
`args->args[i]` may not be NUL-terminated.

We should keep `args->lengths[i]` too.

udf/mrn_udf_highlight_html.cpp
```diff
@@ -287,12 +302,28 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
...
+    } else if (attribute_length == strlen("close_tag") &&
+               strncmp(attribute, "close_tag", strlen("close_tag")) == 0) {
+      info->close_tag = args->args[i];
```
kou, line 320: ditto.

udf/mrn_udf_highlight_html.cpp
```diff
@@ -419,7 +450,9 @@ MRN_API char *mroonga_highlight_html(UDF_INIT *init,
                       reinterpret_cast<grn_pat *>(keywords),
                       args->args[0],
                       args->lengths[0],
-                      &(info->result))) {
+                      &(info->result),
+                      info->open_tag,
+                      info->close_tag)) {
```
kou, line 455: How about passing `info` itself instead?

HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
```diff
@@ -287,12 +302,28 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
...
+    } else if (attribute_length == strlen("close_tag") &&
+               strncmp(attribute, "close_tag", strlen("close_tag")) == 0) {
+      info->close_tag = args->args[i];
+    }
+  }
+
+  if (info->open_tag == nullptr || info->close_tag == nullptr) {
+    info->open_tag = "<span class=\"keyword\">";
+    info->close_tag = "</span>";
   }
```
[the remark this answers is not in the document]
HashidaTKS: Fixed.

Note: It's better to add  a description about this behavior when we add a document for this function.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

HashidaTKS:
> Could you use different text such as `[` not `open_tag` to ensure `open_tag` argument is used?

Fixed

HashidaTKS:
@kou 

Thank you.

udf/mrn_udf_highlight_html.cpp
```diff
@@ -287,12 +304,34 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
...
+    } else if (attribute_length == strlen("close_tag") &&
+               strncmp(attribute, "close_tag", strlen("close_tag")) == 0) {
+      info->close_tag = args->args[i];
+      info->close_tag_length = args->lengths[i];
+    }
+  }
+
+  if (!info->open_tag) {
+    info->open_tag = "<span class=\"keyword\">";
+    info->open_tag_length = strlen("<span class=\"keyword\">");
```
kou, line 330:
```suggestion
    info->open_tag_length = strlen(info->open_tag);
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -287,12 +304,34 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
...
+  }
+  if (!info->close_tag) {
+    info->close_tag = "</span>";
+    info->close_tag_length = strlen("</span>");
```
kou, line 334:
```suggestion
    info->close_tag_length = strlen(info->close_tag);
```

PR #540, opened 2022-09-15
mroonga_highlight_html: add support for html

HashidaTKS:
From https://github.com/mroonga/mroonga/issues/536

Added support for the case of target text is html.
Currently, `mroonga_highlight_html` highlights keywords in tags of html .
Fixed to be able to specify input types (text or html) and not highlight keywords in tags if the input type is html.

Sample SQL
```
SELECT mroonga_highlight_html('<span class="Groonga">Mroonga &amp; Groonga.</span>' AS html,
'Groonga') AS highlighted;
highlighted
<span class="Groonga">Mroonga &amp; <span class="keyword">Groonga</span>.</span>
```

`AS html` at the first argument is important.
If `AS html` is specified, We regard an input text as html.

**Specification when `AS html` is specified**

* Don't escape the HTML special characters like `<` , `>` , `&` and so on.
  * This is because they should be already escaped if the target text is html.
* Don't highlight inside of character references (escaped characters) like `&amp;`.
  * if we specify `&amp;` as target text and `amp` as keyword, the result is `&amp;` and `amp` in `&amp;` is not highlighted.
* Don't highlight inside of HTML tags, means between `<` and `>`, like `<span class="abc">`
  * if we specify `<span class="abc">` as target text and `abc` as keyword, the result is `<span class="abc">` and `abc` in `<>` is not highlighted.
* We can use character references or HTML tags for keywords
  * if we specify `<span class="abc">abc<br/>def</span>` as target text and `c<br/>d` as keyword, the result is `<span class="abc">ab<span class="keyword">c<br/>d</span>ef</span>`.
  * if we specify `<span class="abc">abc&amp;def</span>` as target text and `c&amp;d` as keyword, the result is `<span class="abc">ab<span class="keyword">c&amp;d</span>ef</span>`.

Note that if an user specifies a target text like `<b>abc</b>` and a keyword like `abc</b>`, the result is `<b><span class="keyword">abc</d></span>` and it is broken html.
But it is user's responsibility to specify correct keywords.


HashidaTKS:
@kou @komainu8 

Would please you review this when you have time?

udf/mrn_udf_highlight_html.cpp
```diff
@@ -35,13 +35,19 @@ MRN_BEGIN_DECLS
 extern mrn::DatabaseManager *mrn_db_manager;
 extern mrn::ContextPool *mrn_context_pool;
 
+enum target_type {
```
komainu8, line 38: How about use `enum class` as below?

```c
enum class target_type {
  TEXT = 0,
  HTML
};


```

HashidaTKS: I had fixed this but `enum class` was not supported on centos-7 && mariadb-10.3.
I have reverted the fix...

kou: Could you use `CamelCase` and append `Mrn` prefix?

```suggestion
enum MrnTargetType {
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -35,13 +35,19 @@ MRN_BEGIN_DECLS
...
+  TEXT = 0,
+  HTML = 1
```
kou, line 40: Could you append `MRN_TARGET_TYPE_` prefix? (We can remove them once we can use `enum class`.)

```suggestion
  MRN_TARGET_TYPE_TEXT = 0,
  MRN_TARGET_TYPE_HTML = 1
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -35,13 +35,19 @@ MRN_BEGIN_DECLS
...
+};
+
 typedef struct st_mrn_highlight_html_info
 {
   grn_ctx *ctx;
   grn_obj *db;
   bool use_shared_db;
   grn_obj *keywords;
   grn_obj result;
+  enum target_type target_type;
```
kou, line 50: Could you remove `enum`?

udf/mrn_udf_highlight_html.cpp
```diff
@@ -328,10 +341,24 @@ MRN_API mrn_bool mroonga_highlight_html_init(UDF_INIT *init,
   DBUG_RETURN(true);
 }
 
+static void highlight_html_put_text(grn_ctx *ctx,
+                                       grn_obj *buf,
+                                       const char *str,
+                                       size_t len,
+                                       bool need_escape)
```
kou, line 348:
```suggestion
                                    grn_obj *buf,
                                    const char *str,
                                    size_t len,
                                    bool need_escape)
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -415,12 +445,62 @@ MRN_API char *mroonga_highlight_html(UDF_INIT *init,
...
-                      &(info->result))) {
-    goto error;
+  if (info->target_type == HTML) {
+    const char *previous_position = args->args[0];
+    const char *end_position = args->args[0] + args->lengths[0];
+    const char *current_position = args->args[0];
+    bool is_in_tag = false;
+
+    while (current_position < end_position) {
+      current_position++;
```
kou, line 455: We should not assume a character is one byte.

We can use `grn_charlen()` to count a character size in bytes.

udf/mrn_udf_highlight_html.cpp
```diff
@@ -415,12 +445,62 @@ MRN_API char *mroonga_highlight_html(UDF_INIT *init,
...
+        current_position++;
+        GRN_TEXT_PUT(ctx,
+                     &(info->result),
+                     previous_position,
+                     current_position - previous_position);
+        previous_position = current_position;
+      }
+    }
+
+    if(previous_position < end_position) {
```
kou, line 478:
```suggestion
    if (previous_position < end_position) {
```

udf/mrn_udf_highlight_html.cpp
```diff
@@ -415,12 +445,62 @@ MRN_API char *mroonga_highlight_html(UDF_INIT *init,
...
-                      reinterpret_cast<grn_pat *>(keywords),
-                      args->args[0],
-                      args->lengths[0],
-                      &(info->result))) {
-    goto error;
+  if (info->target_type == HTML) {
+    const char *previous_position = args->args[0];
+    const char *end_position = args->args[0] + args->lengths[0];
+    const char *current_position = args->args[0];
+    bool is_in_tag = false;
```
kou, line 452: `is_` is redundant:

```suggestion
    bool in_tag = false;
```

mysql-test/mroonga/storage/function/highlight_html/t/target_type_html.test
```diff
@@ -0,0 +1,25 @@
...
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+--source ../../../../include/mroonga/not_embedded.inc
+--source ../../../../include/mroonga/have_mroonga.inc
+--source ../../../../include/mroonga/load_mroonga_functions.inc
+
+SELECT mroonga_highlight_html('<span class="Groonga">Mroonga &amp; Groonga.</span>' AS html,
+                              'Groonga') AS highlighted;
```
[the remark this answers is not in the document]
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
```diff
@@ -415,12 +445,69 @@ MRN_API char *mroonga_highlight_html(UDF_INIT *init,
...
+    while (current_position < end_position) {
+      int char_length = grn_charlen(ctx, current_position, end_position);
+
+      if (char_length == 0) {
+        goto error;
+      }
+      
+      current_position += char_length;
+
+      if (*current_position == '<') {
```
kou, line 463: This still assumes the current character is a single byte character.
We need to check the current character byte size is `1` before we access the current character.

HashidaTKS: Fixed.

mysql-test/mroonga/storage/function/highlight_html/t/target_type_html.test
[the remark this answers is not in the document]
HashidaTKS: Fixed.

udf/mrn_udf_highlight_html.cpp
HashidaTKS: Fixed.

HashidaTKS:
@kou @komainu8 

Thank you for your review.

Would you re-review this?

udf/mrn_udf_highlight_html.cpp
```diff
@@ -356,26 +413,41 @@ static bool highlight_html(grn_ctx *ctx,
...
-                              hits[i].offset - previous);
+          highlight_html_put_text(ctx,
+                                  output,
+                                  target + previous,
+                                  hits[i].offset - previous,
+                                  need_escape);
+        }
+        if (skip_tagging_in_char_ref) {
+          in_char_ref = 
+            highlight_html_check_in_char_ref(ctx,
```
kou, line 424: How about checking `&...;` in `mroonga_highlight_html()` too like `<...>`?

HashidaTKS: If we check `&..;` in `mroonga_highlight_html()` and simply skip it, we can not highlight a keyword that contains `&..;` like the following test.
I think we should check whether the current position is in `&..;` after detecting keywords from a target text.

https://github.com/mroonga/mroonga/pull/540/files#diff-9fea1f52acf8445e116584bf15b459fae19c19eeb23a6973e262b3184927f17aR1

HashidaTKS: I think that we can to check whether in the current position is  `<...>` in `highlight_html`, but it may complicate the conditional branching of `highlight_html`. So I decided to check `<...>` in `mroonga_highlight_html()`.

kou: Hmm, what is the expected result for `aaa<b>hello</b>bbb` as the input and `<b>hello</b>` as a keyword?
`aaa<span class="keyword"><b>hello</b></span>bbb`? `aaa<b>hello</b>bbb` (same as the input text) ?

HashidaTKS: In my opinion, users should specify `text` as the input type if a keyword contains tags (`<>`) because `<>` is not appeared inside '<>' and mishighlighting inside `<>` doesn't occur.
But I'm agree with that this is little confusing... Hmm...

HashidaTKS: Ah, I'm wrong, if the input type is `text`, `<` and so on escaped...

HashidaTKS: > Hmm, what is the expected result for `aaa<b>hello</b>bbb` as the input and `<b>hello</b>` as a keyword?
`aaa<span class="keyword"><b>hello</b></span>bbb`? `aaa<b>hello</b>bbb` (same as the input text) ?

I have fixed to get `aaa<span class="keyword"><b>hello</b></span>bbb` when a user specify `aaa<b>hello</b>bbb` as the input and `<b>hello</b>` as a keyword.

kou: Ah, if an user specifies `<b>hello` as a keyword, it will generate a broken HTML...

HashidaTKS: Yes, but I think it is an user's responsibility to specify a correct keyword.
How do you think to just adding a notification about this for the `mroonga_highlight_html` document for now and fixing it when we need to fix it in the future?

kou: Could you summarize the current specification? I want to confirm whether there is inconsistency or not.

HashidaTKS: > Could you summarize the current specification? I want to confirm whether there is inconsistency or not.

I have added the current specification into the description of this PR.
https://github.com/mroonga/mroonga/pull/540#issue-1374262373

Please confirm it.

PRO-BROS:
HI!

plz take a time and review this [site](https://prhocoding.com/)

udf/mrn_udf_highlight_html.cpp
kou: Considering use cases for this feature, the following specification seems hard to use:

> * We can use character references or HTML tags for keywords
>   * if we specify `<span class="abc">abc<br/>def</span>` as target text and `c<br/>d` as keyword, the result is `<span class="abc">ab<span class="keyword">c<br/>d</span>ef</span>`.
>   * if we specify `<span class="abc">abc&amp;def</span>` as target text and `c&amp;d` as keyword, the result is `<span class="abc">ab<span class="keyword">c&amp;d</span>ef</span>`.
>
> Note that if an user specifies a target text like `<b>abc</b>` and a keyword like `abc</b>`, the result is `<b><span class="keyword">abc</d></span>` and it is broken html.
> But it is user's responsibility to specify correct keywords.

An use case (see #536):
* Extract HTML tags from target texts manually (in an app) and put the extracted text to Groonga
* Search an input from end-users (not app developers) against the extracted text
* Use the input from end-users (not app developers) to highlight the original HTML included text (not the extracted text)

In this use case, the target extracted text doesn't have HTML tags. So `c<br/>d` isn't matched. So it should not be highlighted.

In this use case, keywords for this context are specified by end-users not app developers. So "it is user's responsibility to specify correct keywords" is hard. Because app developers can't control it.

How about the followings?

* We can use HTML tags for keywords but they are never matched
* We can use character references (bonus)
  * We expand character references in target text and keyword before we process: `&amp;` -> `&`
  * We highlight with the expanded target text and the expanded keyword, and then escape the result like `AS text` case:
    * target text: `a &amp; b &#38; c`
      * keyword: `&` -> `a <span class="keyword">&amp;</span> b <span class="keyword">&amp;<span> c`
      * keyword: `&amp;` -> `a <span class="keyword">&amp;</span> b <span class="keyword">&amp;<span> c`
* Highlighted HTML isn't broken if the target text isn't broken

HashidaTKS: Thanks.

>  * We expand character references in target text and keyword before we process: `&amp;` -> `&`

I think we need to support all of character references for it.

https://html.spec.whatwg.org/multipage/named-characters.html

Are there good libraries or functions for it?

kou: We can use them statically because it'll not be changed:

> This list is static and [will not be expanded or changed in the future](https://github.com/whatwg/html/blob/main/FAQ.md#html-should-add-more-named-character-references).

I think that using an external library is overkill for our use case.

HashidaTKS:
A long time has passed since this PR was created.

I think we should re-consider this again from the beginning.
We can discuss about this in https://github.com/mroonga/mroonga/issues/536

PR #542, opened 2022-09-20
packages mariadb: bump mariadb versions

HashidaTKS:


komainu8:
Could you version up release version of Mroonga package like https://github.com/mroonga/mroonga/commit/21bc19685d7605210a9f4d1b1527149ae638b640#diff-42eb71b6a4a0256fd04202798dccadf2bd5e9d0330ea32f6b9870b3230ea89f1L27.

komainu8:
Could you add change log like https://github.com/mroonga/mroonga/commit/21bc19685d7605210a9f4d1b1527149ae638b640#diff-42eb71b6a4a0256fd04202798dccadf2bd5e9d0330ea32f6b9870b3230ea89f1R130-R133?

HashidaTKS:
@komainu8 

Thank you for your review!

> If we don't update the release version of the package, Mroonga users can't use the new package. Because if the release version is same new package and old package, yum or dnf can't detect new package.

Please let me confirm my understanding about this.
When MariaDB version is updated for a certain Mroonga version, we need update the release version in these `spec.in` files.
When Mroonga version is updated, the release version is reset to 1.
Is it right?

komainu8:
> When Mroonga version is updated, the release version is reset to 1. Is it right?

Yes, it is.
The release version of Mroonga package is reset when Mroonga releases.

HashidaTKS:
> Yes, it is.
> The release version of Mroonga package is reset when Mroonga releases.

I see. Thanks!

HashidaTKS:
@komainu8 

May I merge this?

HashidaTKS:
Thanks!

PR #559, opened 2022-10-27
Fix max_execution_time test

HashidaTKS:
`MAX_EXECUTION_TIME` didn't throw any error because of the MySQL's `sleep()` function specification.

https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_sleep

> When [SLEEP()](https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_sleep) is the only thing invoked by a query that is interrupted, it returns 1 and the query itself returns no error. This is true whether the query is killed or times out:

The current query seems to apply to this case.

> When [SLEEP()](https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_sleep) is only part of a query that is interrupted, the query returns an error:

I have changed the test query into like that.


mysql-test/mroonga/storage/r/max_execution_time.result
```diff
@@ -4,12 +4,12 @@ CREATE TABLE diaries(
...
-SELECT sleep(1), title FROM diaries;
-sleep(1)	title
-0	Hello
-SELECT /*+ MAX_EXECUTION_TIME(10000) */ sleep(1), title FROM diaries;
-sleep(1)	title
-0	Hello
+SELECT title FROM diaries WHERE sleep(1) OR title IS NOT NULL;
+title
+Hello
+SELECT /*+ MAX_EXECUTION_TIME(10000) */ title FROM diaries WHERE sleep(1) OR title IS NOT NULL;
```
HashidaTKS, line 12: `OR title IS NOT NULL` is for getting values of the title column.
If we specify only `sleep()`, there is no match records.

kou: Could you add it to `.test` as a comment?

kou (commented):
Good catch!

(As far as I remember, the test (only `sleep()`) worked with `ENGINE=InnoDB` but didn't work with `ENGINE=Mroonga`...)

This test sill failed with MySQL 5.7:

https://github.com/mroonga/mroonga/actions/runs/3339284180/jobs/5533718527#step:5:1464

```text
[ 41%] plugin/mroonga/storage.max_execution_time w1 [ fail ]
        Test ended at 2022-10-27 17:57:03

CURRENT_TEST: plugin/mroonga/storage.max_execution_time
mysqltest: At line 35: query 'SELECT /*+ MAX_EXECUTION_TIME(1) */ title FROM diaries WHERE sleep(1) OR title IS NOT NULL' succeeded - should have failed with errno 3024...

The result from queries just before the failure was:
DROP TABLE IF EXISTS diaries;
SET NAMES utf8mb4;
CREATE TABLE diaries(
title TEXT
) DEFAULT CHARSET=utf8mb4;
INSERT INTO diaries VALUES ('Hello');
safe_process[4216]: Child process: 4217, exit: 1
```

Is this another reason? Should we still disable this for MySQL 5.7?

HashidaTKS:
@kou 

Thanks.

> This test sill failed with MySQL 5.7:

> Is this another reason? Should we still disable this for MySQL 5.7?

I will check this.

PR #567, opened 2022-11-09
test: update expected values of "SHOW CREATE TABLE"

komainu8:
Because "SHOW CREATE TABLE" display default collations since 10.3.37, 10.4.27, 10.5.18, 10.6.11, 10.7.7, 10.8.6, 10.9.4.
See: https://jira.mariadb.org/browse/MDEV-29446.

The modification of https://jira.mariadb.org/browse/MDEV-29446 does not include MySQL and Percona Server.


komainu8:
@kou @HashidaTKS Could you review this PR?

kou:
We don't want to add tests for this difference to reduce maintenance cost.

Can we remove the default collate by `--replace_regex / COLLATE=utf8mb4_general_ci//` or something?

kou:
Done.

PR #575, opened 2022-11-28
packages: add support for MariaDB 10.10

komainu8:
MariaDB 10.10 changed the default value of `explicit_defaults_for_timestamp` to `ON`.
* https://jira.mariadb.org/browse/MDEV-28632


mysql-test/mroonga/include/mroonga/check_version.inc
```diff
@@ -15,28 +15,32 @@
...
-let $version_5_5_or_later  = `SELECT $version_major_minor >= 5.5`;
-let $version_5_6_or_later  = `SELECT $version_major_minor >= 5.6`;
-let $version_5_7_or_later  = `SELECT $version_major_minor >= 5.7`;
-let $version_8_0_or_later  = `SELECT $version_major_minor >= 8.0`;
-let $version_10_0_or_later = `SELECT $version_major_minor >= 10.0`;
-let $version_10_2_or_later = `SELECT $version_major_minor >= 10.2`;
-let $version_10_3_or_later = `SELECT $version_major_minor >= 10.3`;
-let $version_10_4_or_later = `SELECT $version_major_minor >= 10.4`;
-let $version_10_5_or_later = `SELECT $version_major_minor >= 10.5`;
+let $version_5_5_or_later  = `SELECT $version_major_minor >= 505`;
```
HashidaTKS, line 37: Or using the following query and not defining `$version_major_minor`.

```
SELECT $version_major = 5 OR ($version_major = 5 AND $version_minor = 5)
```

mysql-test/mroonga/include/mroonga/check_version.inc
```diff
@@ -15,28 +15,32 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 
 --disable_query_log
-let $version_major_minor =
-  `SELECT CAST(SUBSTRING_INDEX(@@global.version, '.', 2) AS DECIMAL(4, 2))`;
+let $version_major =
+  `SELECT CAST(SUBSTRING_INDEX(@@global.version, '.', 1) AS UNSIGNED)`;
+let $version_minor =
```
HashidaTKS, line 20: I think this is not elegant.
Is there other good way to get major/minor versions?

mysql-test/mroonga/include/mroonga/check_version.inc
```diff
@@ -15,28 +15,32 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 
 --disable_query_log
-let $version_major_minor =
-  `SELECT CAST(SUBSTRING_INDEX(@@global.version, '.', 2) AS DECIMAL(4, 2))`;
```
kou, line 19: Why do we need these changes?
Is it for distinguishing `10.1` and `10.10`? Then how about the following?

```sql
SELECT CAST(
         CONCAT(
           SUBSTRING_INDEX(@@global.version, '.', 1),
           '.',
           LPAD(
             SUBSTRING_INDEX(
               SUBSTRING_INDEX(@@global.version, '.', 2),
               '.',
               -1
             ),
             2,
             '0'
           )
         )
         AS DECIMAL(4, 2)
       );
```

HashidaTKS: > Is it for distinguishing 10.1 and 10.10? Then how about the following?

Yes, currently 10.10 is regarded as 10.1, and tests failed by that reason.
Tests that should be skipped greater or equals to 10.5 are executed because 10.10 is regarded as 10.1.

HashidaTKS: I will try this SQL!

komainu8:
Reference: https://jira.mariadb.org/browse/MDEV-28632

HashidaTKS:
> Reference: https://jira.mariadb.org/browse/MDEV-28632

Thanks, the release note for MariaDB 10.10 also refers to it.
https://mariadb.com/kb/en/mariadb-10100-release-notes/#miscellaneous

mysql-test/mroonga/include/mroonga/have_mroonga.inc
```diff
@@ -44,4 +44,10 @@ if ($have_default_tmp_storage_engine_variable) {
   set default_tmp_storage_engine=Mroonga;
 }
 
+if ($mariadb) {
+  if ($version_10_10_or_later) {
+    SET explicit_defaults_for_timestamp=false;
```
HashidaTKS, line 49: The default value of `explicit_defaults_for_timestamp` is changed false to true since 10.10 in MariaDB.
Some tests fail if `explicit_defaults_for_timestamp` is true because a displayed type is different from the expected results.
In order to avoid that failing, `explicit_defaults_for_timestamp` are set to false.

But I doubt that it is good to set `explicit_defaults_for_timestamp=false` in this file.
The reason why I added it in this file is because this file is executed on all test cases, 
but I feel it is not good in semantic.
(`have_mroonga` should do about Mroonga, but `explicit_defaults_for_timestamp` is the parameter for DBMS, not Mroonga)

Is there any good place to set `explicit_defaults_for_timestamp` to false?

komainu8: How about mtr's option as below.

https://github.com/mroonga/mroonga/blob/master/packages/yum/test.sh#L191

```
--mysqld=--explicit_defaults_for_timestamp=ON
```

BTW, I think that we should `explicit_defaults_for_timestamp` is ON.
Because TIMESTAMP with implicit DEFAULT value is deprecated.
Therefore, in generally, we should set the value of `explicit_defaults_for_timestamp` is ON.

The default value of `explicit_defaults_for_timestamp` in MySQL 8.0.
https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_explicit_defaults_for_timestamp

HashidaTKS: > Are these 3 tests depend on the default value? 

No, just not specifying DEFAULT.

> How about just adding `DEFAULT CURRENT_TIMESTAMP` explicitly to these 3 tests?

So, I have adopted this suggest.

HashidaTKS:
@komainu8 @kou 

PR #583, opened 2022-12-15
doc release: Add description about Docker

HashidaTKS:


doc/source/developer/release.rst
```diff
@@ -208,6 +208,32 @@ Windows
...
+--------------------
+
+We update Mroonga's Docker images of `Docker Hub <https://hub.docker.com/r/groonga/mroonga>`_ .
+
+Clone `the Mroonga's Docker repository <https://github.com/mroonga/docker>`_ and update Dockerfiles in tha repository.
+
+Here is an example for the case that the MySQL version is ``5.7.26``, the Mroonga version is ``9.01``, the Groonga version is ``9.0.2``::
+
+    % mkdir -p ~/work/mroonga
+    % rm -rf ~/work/mroonga/docker.clean
```
kou, line 221: Should we always remove a previously cloned repository?

HashidaTKS: Strictly no, but it is recommended to create `groonga.clean`, `mroonga.clean` and so on for each releases to avoid containing unintended changes.
https://groonga.org/ja/docs/contribution/development/release.html#id7

Following those, I think it is better to create `docker.clean` for each releases.

kou: How about running `git clean -xdf` when we reuse the cloned repository instead?

HashidaTKS: Okay, I have modified the code sample to run `git clean -fdx`, `git checkout .`, `git pull`

PR #584, opened 2022-12-20
package mariadb-10.5 bullseye: rebuild against the latest MariaDB 10.5 package

komainu8:
We can marge this PR when this PR pass all CI.


HashidaTKS:
The failed test will be fixed after we update the current package for mariadb-10.5 bullseye.

PR #596, opened 2023-01-26
yum test: enable upgrade test for MariaDB 10.10

komainu8:
Because we released Mroonga 12.10.


komainu8:
I'll merge this PR when all CI pass.

Issue #605, opened 2023-02-17
 Error compiling the mroonga plugin 13.00 with MariaDB 10.11.2 in Debian 11.6


josepsanzcamp:
### What happend?

Hi.

I have updated my MariaDB to the latest release 10.11.2 and I get this error when I try to compile the mroonga plugin:

```
ha_mroonga.cpp: In member function ‘bool ha_mroonga::storage_inplace_alter_table_add_column(TABLE*, Alter_inplace_info*)’:
mrn_mysql_compat.h:709:42: error: no matching function for call to ‘TABLE::update_virtual_field(Field*&)’
  709 |        (table->update_virtual_field(field))
      |                                          ^
ha_mroonga.cpp:17370:9: note: in expansion of macro ‘MRN_GENERATED_COLUMNS_UPDATE_VIRTUAL_FIELD’
17370 |         MRN_GENERATED_COLUMNS_UPDATE_VIRTUAL_FIELD(altered_table, altered_field);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from mroonga-13.00/../mariadb-10.11.2+maria~deb11//sql/field.h:29,
                 from mroonga-13.00/../mariadb-10.11.2+maria~deb11//sql/sql_class.h:31,
                 from mrn_mysql.h:57,
                 from ha_mroonga.cpp:24:
mroonga-13.00/../mariadb-10.11.2+maria~deb11//sql/table.h:1737:7: note: candidate: ‘int TABLE::update_virtual_field(Field*, bool)’
 1737 |   int update_virtual_field(Field *vf, bool ignore_warnings);
      |       ^~~~~~~~~~~~~~~~~~~~
mroonga-13.00/../mariadb-10.11.2+maria~deb11//sql/table.h:1737:7: note:   candidate expects 2 arguments, 1 provided
make[2]: *** [Makefile:652: ha_mroonga_la-ha_mroonga.lo] Error 1
make[2]: Leaving directory 'mroonga-13.00'
make[1]: *** [Makefile:681: all-recursive] Error 1
make[1]: Leaving directory 'mroonga-13.00'
make: *** [Makefile:512: all] Error 2
```

Could you help me to solve this issue???

Thanks in advance.

Josep.

### How to reproduce it

By compilint it

### Expected behavior

No error while compile

### Environment

- OS: Debian
- OS Version: Debian 11.6
- DB: MariaDB
- DB Version: 10.11.2
[...]

PR #606, opened 2023-02-17
Add support for building with MariaDB 10.11.2

josepsanzcamp:
GitHub: fix GH-605 

Error compiling the Mroonga plugin 13.00 with MariaDB 10.11.2 in Debian 11.6.

### DIFF
--- a/mrn_mysql_compat.h
+++ b/mrn_mysql_compat.h
@@ -694,7 +694,8 @@ typedef uint mrn_srid;
 #endif
 
 #ifdef MRN_MARIADB_P
-#  if (((MYSQL_VERSION_ID >= 101002) && (MYSQL_VERSION_ID < 101100)) || \
+#  if (((MYSQL_VERSION_ID >= 101102) && (MYSQL_VERSION_ID < 101200)) || \
+       ((MYSQL_VERSION_ID >= 101002) && (MYSQL_VERSION_ID < 101100)) || \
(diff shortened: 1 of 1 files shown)


mrn_mysql_compat.h
```diff
@@ -694,7 +694,8 @@ typedef uint mrn_srid;
 #endif
 
 #ifdef MRN_MARIADB_P
-#  if (((MYSQL_VERSION_ID >= 101002) && (MYSQL_VERSION_ID < 101100)) || \
+#  if (((MYSQL_VERSION_ID >= 101102) && (MYSQL_VERSION_ID < 101200)) || \
+       ((MYSQL_VERSION_ID >= 101002) && (MYSQL_VERSION_ID < 101100)) || \
```
kou, line 698: How about removing `< 101100` check instead? We will not support MariaDB 10.11.1 because MariaDB 10.11.1 is RC not GA: https://mariadb.com/kb/en/changes-improvements-in-mariadb-1011/#list-of-all-mariadb-1011-releases

```suggestion
#  if (((MYSQL_VERSION_ID >= 101002) || \
```

josepsanzcamp: Hi @kou, your suggestion have an extra parenthesis and causes an error when compile.

kou: Oh, sorry.

kou:
Thanks!

PR #608, opened 2023-02-20
packages percona-server: enable build of packages for Percona Server

komainu8:
Because the library to need build the RPM package of Percona Server already has been able to aveirable for download.


komainu8:
I'll merge this pull request when all CI passed.

PR #620, opened 2023-04-06
packages: drop support for Ubuntu 18.04

komainu8:
Because Ubuntu 18.04 reached EOL.


packages/mysql-5.7-mroonga/Rakefile
kou, line 1: Can we remove this directory?

komainu8: Thanks.
I've fixed it.

kou: Could you also update https://github.com/mroonga/mroonga/blob/master/packages/Rakefile#L11 ?

komainu8: Ah...Sorry.
I've fixed it

kou: We may want to run `cd packages && rake version:update` in a CI job to detect this case automatically.

kou (approved):
+1

komainu8:
Thank you for your comments!

Issue #632, opened 2023-06-15
Add a newly support for MariaDB 11.0.2


komainu8:
### What is your problem?

The first GA release of the MariaDB 11.0.2 at 2023-06-06.

See: https://mariadb.org/mariadb-11-1-1-11-0-2-now-available/

komainu8:
I close this issue.
Because MariaDB 11.x.x is short-term release.

PR #633, opened 2023-06-15
packages mariadb: add a newly support for MariaDB 11.0

komainu8:
Fix: #632

* Use `keyread_time()` + `rnd_pos_time()` instead of `read_time()`.
  Because `read_time()` is removed since MariaDB 11.0.2.

* Change the type of return value to `IO_AND_CPU_COST` for `scan_time()`.
  Because the return value of type of `scan_time()` is changed `IO_AND_CPU_COST` since MariaDB 11.0.2.

See:
* https://github.com/MariaDB/server/commit/b66cdbd1eaeed7e96317a03a190c496fd062ec71

---

* Add a newly argument of range optimizer.
  Because the argument of `multi_range_read_info_const()` is added since MariaDB11.0.2.

See:
* https://github.com/MariaDB/server/commit/009db2288ba8d22c5e8e5e047d8f8694f7097923

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -230,6 +230,17 @@ jobs:
             package-type: apt
             package: mariadb-10.11
 
+          # MariaDB 11.0
+          - os: almalinux-8
+            package-type: yum
+            package: mariadb-11.0
+          - os: almalinux-9
+            package-type: yum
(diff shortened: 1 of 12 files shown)


komainu8:
I close this PR.
Because MariaDB 11.x.x is short-term release.

PR #636, opened 2023-07-20
test: skip tests for changing binlog format in MySQL 8.0 or later

komainu8:
Because "binlog_format" parameter is deprecated as of MySQL 8.0.34.

See:
https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_binlog_format


kou:
We should just skip instead of removing because MariaDB still has the parameter.

mysql-test/mroonga/storage/update/t/binlog_row_for_mysql_8_0_or_later.test
kou, line 1: Do we need this?
What is tested by this?

komainu8: This test is for confirming that we can execute "UPDATE" against normal columns that we don't set index even if Mroonga storage mode.

Just in case, I added this test.

kou: Then we should not use `binlog_row` for filename.
The existing `int.test` doesn't cover the case?

komainu8: Ah...I see.
`int.test` can cover the case.

So, I remove `mysql-test/mroonga/storage/update/t/binlog_row_for_mysql_8_0_or_later.test`.

komainu8: I've removed it.

PR #641, opened 2023-07-27
doc news: separate translation files

komainu8:


kou:
We don't need to add `doc/locale/en/LC_MESSAGES/**/*.po` because we never change them.

PR #655, opened 2023-10-13
ci windows: add support ASM language

komainu8:
This modification is trial.
I’m not sure if this modification will work yet.


komainu8:
I don't understand this problem...

PR #659, opened 2023-10-23
rpm: don't use "which" command

komainu8:
Because LXD images of Almalinux 8,9 and CentOS 7 have not "which" command in default.

### DIFF
--- a/packages/yum/test.sh
+++ b/packages/yum/test.sh
@@ -33,6 +33,7 @@ case ${major_version} in
     ;;
 esac
 
+sudo ${DNF} install -y which
 sudo ${DNF} install -y \
   https://packages.groonga.org/${os}/${major_version}/groonga-release-latest.noarch.rpm
 
--- a/packages/yum/test.sh
+++ b/packages/yum/test.sh
@@ -33,7 +33,6 @@ case ${major_version} in
     ;;
 esac
 
-sudo ${DNF} install -y which
 sudo ${DNF} install -y \
   https://packages.groonga.org/${os}/${major_version}/groonga-release-latest.noarch.rpm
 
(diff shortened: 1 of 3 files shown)


packages/yum/test.sh
```diff
@@ -33,6 +33,7 @@ case ${major_version} in
     ;;
 esac
 
+sudo ${DNF} install -y which
```
kou, line 36: We should add `Requires: which` to `.spec` because it's used in package not test script.

Or we may be able to just stop using `which`:

```diff
diff --git a/data/rpm/post.sh b/data/rpm/post.sh
index ac561df0..d5da9331 100755
--- a/data/rpm/post.sh
+++ b/data/rpm/post.sh
@@ -69,7 +69,7 @@ update_sql=${data_dir}/mroonga/update.sql
 
 need_password_expire=no
 if [ "${try_auto_prepare}" = "yes" ]; then
-  mysql_command=$(which mysql)
+  mysql_command=mysql
   password_option=""
   if [ "${have_auto_generated_password}" = "yes" ]; then
     if "${mysql_command}" \
diff --git a/data/rpm/preun.sh b/data/rpm/preun.sh
index dedde0f1..21e285c8 100755
--- a/data/rpm/preun.sh
+++ b/data/rpm/preun.sh
@@ -64,7 +64,7 @@ uninstall_sql=${data_dir}/mroonga/uninstall.sql
 
 need_password_expire=no
 if [ "${try_auto_uninstall}" = "yes" ]; then
-  mysql_command=$(which mysql)
+  mysql_command=mysql
   password_option=""
   if [ "${have_auto_generated_password}" = "yes" ]; then
     if "${mysql_command}" \
```

komainu8: Ah...I see.

I will try to stop using `which`.

PR #660, opened 2023-11-02
ci windows: remove needless expression

komainu8:
Because we have already released groonga-normalizer-mysql 1.2.3.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -164,7 +164,7 @@ jobs:
           submodules: recursive
       - name: Download groonga-normalizer-mysql latest
         if: |
-          !(matrix.use-main == 'yes')
+          matrix.use-main != 'yes'
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -156,19 +156,15 @@ jobs:
...
         with:
           path: mariadb/extra/groonga-normalizer-mysql
           repository: groonga/groonga-normalizer-mysql
           submodules: recursive
       - name: Download groonga-normalizer-mysql latest
-        # TODO: Remove " -Or $true" when we release
-        # groonga-normalizer-mysql 1.2.3.
         if: |
-          !(matrix.use-main == 'yes' || true)
+          !(matrix.use-main == 'yes')
```
kou, line 167:
```suggestion
          matrix.use-main != 'yes'
```

PR #666, opened 2024-01-19
packages mysql-community-8.0: use the latest settings of MySQL's repositories for CentOS7, Debian bullseye, Debian bookworm, and Ubuntu jammy

komainu8:
These modifications same as #663.

I forgot modification for CentOS7, Debian bullseye, Debian bookworm, and Ubuntu jammy.


komainu8:
GPG error of CentOS 7 resolved: https://github.com/mroonga/mroonga/actions/runs/7578326725/job/20640927928
GPG error of Debian bullseye resolved: https://github.com/mroonga/mroonga/actions/runs/7578326725/job/20640928073
GPG error of Ubuntu jammy resolved: https://github.com/mroonga/mroonga/actions/runs/7578326725/job/20640927496 

The result of Debian bookworm is none.
Because we don't create the package for Debian bookworm yet.

PR #667, opened 2024-01-19
packages mysql-community-server-8.0: enable package for Debian bookworm

komainu8:
Because the source of MySQL Community Server for Debian 12 was released in MySQL APT Repository(https://repo.mysql.com/apt/debian/dists/).

---

I am trying whether CI pass or not.


kou:
Done.

PR #670, opened 2024-02-19
Rebuilt against MySQL 8.0.35

abetomo:
* https://launchpad.net/~groonga/+archive/ubuntu/ppa/+sourcepub/15799241/+listing-archive-extra
* https://launchpad.net/~groonga/+archive/ubuntu/ppa/+sourcepub/15799242/+listing-archive-extra

### DIFF
--- a/packages/mysql-8.0-mroonga/debian/changelog
+++ b/packages/mysql-8.0-mroonga/debian/changelog
@@ -1,3 +1,9 @@
+mysql-8.0-mroonga (13.05-3) unstable; urgency=low
+
+  * Rebuilt against MySQL 8.0.35.
+
+ -- Abe Tomoaki <abe@clear-code.com>  Fri, 16 Feb 2024 05:45:42 -0000
+
 mysql-8.0-mroonga (13.05-2) unstable; urgency=low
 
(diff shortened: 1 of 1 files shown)


packages/mysql-8.0-mroonga/debian/changelog
```diff
@@ -1,3 +1,9 @@
+mysql-8.0-mroonga (13.05-3) unstable; urgency=low
+
+  * Rebuilt against MySQL 8.0.35.
```
komainu8, line 3: mysql-8.0-mroonga (13.05-3) is built with MySQL 8.0.36.
So, we should write "Rebuilt against MySQL 8.0.36" in this change log.

kou: Good catch!

In general, we should generate the message automatically to prevent this type mistake.

abetomo:
```
mysql --version
mysql  Ver 8.0.36-0ubuntu0.22.04.1 for Linux on x86_64 ((Ubuntu))
```

Oh. That's for sure.

PR #683, opened 2024-05-28
packages mysql-community-minimal-8.0: install mysql-community-release before mysql-community-minimal-release

komainu8:
TODO: Test

If we install mysql-community-release after mysql-community-minimal-release, we can't install mysql-community-minimal-release.

Because mysql-community-minimal-release and mysql80-community-release are conflicting. 
See: https://github.com/mroonga/mroonga/actions/runs/9265261501/job/25487027793#step:9:5291

However, if we install mysql-community-release firstly, we can replace mysql80-community-release to mysql-community-minimal-release.

Currently, We install mysql-community-release only to install mysql-community-devel in this job. So, no problem that mysql-community-release replace to mysql-community-minimal-release, anytime after we install mysql-community-devel.


kou:
mysql-community-minimal-8.0ってまだ必要なんだっけ？
これ、Docker用だよね？

PR #685, opened 2024-05-31
ci test: use the latest mysql-apt-config

komainu8:
Currently, we couldn't be verified signatures of 'http://repo.mysql.com/apt/debian bookworm InRelease' as below.
https://github.com/mroonga/mroonga/actions/runs/9309443050/job/25625429475#step:6:781

Because we use old public key.

### DIFF
--- a/packages/apt/test.sh
+++ b/packages/apt/test.sh
@@ -78,7 +78,7 @@ case ${package} in
     ;;
   mysql-community-*)
     old_package=${package}
-    wget https://repo.mysql.com/mysql-apt-config_0.8.24-1_all.deb
+    wget https://repo.mysql.com/mysql-apt-config_0.8.30-1_all.deb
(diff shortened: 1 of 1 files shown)


packages/apt/test.sh
```diff
@@ -78,7 +78,7 @@ case ${package} in
     ;;
   mysql-community-*)
     old_package=${package}
-    wget https://repo.mysql.com/mysql-apt-config_0.8.24-1_all.deb
+    wget https://repo.mysql.com/mysql-apt-config_0.8.30-1_all.deb
```
kou, line 81:
```suggestion
    wget https://repo.mysql.com/mysql-apt-config.deb
```

komainu8: Ah..Thank you for your comment.

PR #711, opened 2024-06-19
Implement "make tag" by Ruby

komainu8:
We will drop support for GNU Autotools. This is a part of the work. We can use `rake release:tag` instead of `make tag`.

The behavior of this modification https://github.com/mroonga/mroonga/blob/v13.05/Makefile.am#L45-L47 about the same as below.

* Setting tag by `git tag vXX.XX -a -m "Mroonga XX.XX!!!"`
* Push tag to Mroonga repository by `git push origin "vXX.XX"`

Usage:

$ rake release:tag

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -74,6 +74,17 @@ namespace :release do
          "package: update version info to #{version} (#{new_release_date})")
     end
   end
+
+  desc "Tag"
+  task :tag do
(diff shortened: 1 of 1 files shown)


otegami (commented):
LGTM 👍🏾

PR #733, opened 2024-07-03
ci: use upload-artifact@v3 temporarily

komainu8:
Because if we use upload artifact by using upload-artifact@v3, we can not download artifact on branch by using REST API.

So, we use upload-artifact@v3 temporarily.
Because this problem doesn't occur in upload-artifact@v3.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -82,7 +82,7 @@ jobs:
           make dist
 
       # Artifact
-      - uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


kou:
これって、`main`じゃなくて`maintenance/*`だけで必要じゃない？

komainu8:
> これって、`main`じゃなくて`maintenance/*`だけで必要じゃない？

あぁ、そうですね。
`maintenance/*`にだけ入れるようにします。

PR #741, opened 2024-07-25
packages mysql-community-8.0: use the latest MySQL Community Server

komainu8:


komainu8:
Upgrade Test fail yet, but this test can't fix as long as we don't upload rebuilding package.
So, I merge this PR.

PR #790, opened 2024-10-03
{tiny,small,}int: update index for `NULL` with `0` to support `index_column_diff`

otegami:
GitHub: GH-789

The current implementation ignores `NULL`. It means that we don't update index for `NULL`. But it causes `index_column_diff` false positive.

`{tiny,small,}int`'s `NULL` are processed as `0` in Groonga. Because Groonga uses the default value for `NULL`. `index_column_diff` uses `0` for `NULL` in the expected posting lists. It causes false positive.

If we use `0` instead of ignoring for `NULL`, we can avoid the false positive. It also enables that we can search `NULL` column value record by `0`. In general, it'll be useful but this is an incompatible change. But it'll be acceptable because `NULL` behavior is undefined by definition.


mysql-test/mroonga/storage/column/int/t/with_index_null_value.test
```diff
@@ -0,0 +1,42 @@
+# Copyright(C) 2024 Takuya Kodama <otegami@clear-code.com>
```
kou, line 1: We should use this format: https://github.com/mroonga/mroonga/blob/56052c854466420ba4df39847561bbccd5993838/COPYING#L474

```suggestion
# Copyright (C) 2024  Takuya Kodama <otegami@clear-code.com>
```

(Old tests don't use the format...)

The following mode line is better for Emacs users like me:

```suggestion
# -*- mode: sql; sql-product: mysql -*-
#
# Copyright (C) 2024  Takuya Kodama <otegami@clear-code.com>
```

See also: https://github.com/mroonga/mroonga/blob/56052c854466420ba4df39847561bbccd5993838/mysql-test/mroonga/storage/index/desc/t/mariadb.test#L1

mysql-test/mroonga/storage/column/int/t/with_index_null_value.test
kou, line 1: Could you use `null.test` for file name?

mysql-test/mroonga/storage/column/int/t/with_index_null_value.test
```diff
@@ -0,0 +1,42 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+--source ../../../../include/mroonga/have_mroonga.inc
+--source ../../../include/mroonga/load_mroonga_functions.inc
```
kou, line 18: The number of `../` are different.

mysql-test/mroonga/storage/column/int/t/with_index_null_value.test
```diff
@@ -0,0 +1,42 @@
...
+
+INSERT INTO items VALUES ("hamburger", 200);
+INSERT INTO items VALUES ("smile", 0);
+INSERT INTO items VALUES ("sample chips", NULL);
+
+SELECT mroonga_command("index_column_diff --table items#price_index --name index");
+
+SELECT * FROM items WHERE price = 0;
+
+SELECT * FROM items WHERE price >= 100;
```
kou, line 38: This is needless.

mysql-test/mroonga/storage/column/int/t/with_index_null_value.test
otegami: fix: [f66331b](https://github.com/mroonga/mroonga/pull/790/commits/f66331b3dbec6e4643bd3a9cfdf0e945e5f55fda)

mysql-test/mroonga/storage/column/int/t/with_index_null_value.test
otegami: fix: [08c15be](https://github.com/mroonga/mroonga/pull/790/commits/08c15bee929323a3d42e00047e2ecefc08be2a4f)

mysql-test/mroonga/storage/column/int/t/with_index_null_value.test
otegami: fix: [800d984](https://github.com/mroonga/mroonga/pull/790/commits/800d9847ee518059275e7418ae73e44cabfab006)

mysql-test/mroonga/storage/column/int/t/with_index_null_value.test
otegami: fix: [c3b9abc](https://github.com/mroonga/mroonga/pull/790/commits/c3b9abc9cfcc561ebda54e4a6cecf326a0409af8)

mysql-test/mroonga/storage/column/int/t/null.test
```diff
@@ -0,0 +1,41 @@
...
+INSERT INTO items VALUES ("smile", 0);
+INSERT INTO items VALUES ("sample chips", NULL);
+
+SELECT mroonga_command("index_column_diff --table items#price_index --name index");
+
+SELECT * FROM items WHERE price = 0;
+
+DROP TABLE items;
+
+--source ../../../../include/mroonga/have_mroonga_deinit.inc
```
kou, line 41:
```suggestion
--source ../../../../include/mroonga/unload_mroonga_functions.inc
--source ../../../../include/mroonga/have_mroonga_deinit.inc
```

otegami: fix: [98bfd6c](https://github.com/mroonga/mroonga/pull/790/commits/98bfd6c96a77c761fcc80b177ef97e7a7f28f66f) Thanks you so much!

ha_mroonga.cpp
```diff
@@ -7324,7 +7324,12 @@ int ha_mroonga::storage_write_row(mrn_write_row_buf_t buf)
   for (i = 0; i < n_columns; i++) {
     Field *field = table->field[i];
 
-    if (field->is_null())
+    // TODO: Remove this when support for handling how to register NULLs
+    //       in the index is implemented for all columns.
+    if (field->is_null() &&
+        ((field->real_type() != MYSQL_TYPE_TINY) &&
+         (field->real_type() != MYSQL_TYPE_SHORT) &&
+         (field->real_type() != MYSQL_TYPE_LONG)))
```
kou, line 7332: `MYSQL_TYPE_LONG`だけにするか`tinyint`と`smallint`のテストも追加してもらえますか？

otegami: 98bfd6c...cbc618d にて、`tinyint`と`smallint`のテストの追加を行いました 🙏🏾

otegami:
ここまで頂いたレビューコメントの対応をおこないました 🙏🏾

PR #791, opened 2024-10-07
time fraction seconds: update index for NULL with `00:00:00.0` to support `index_column_diff`

otegami:
GitHub: GH-789

The current implementation ignores NULL. It means that we don't update index for NULL.

`time` with fraction seconds' NULL is processed as `00:00:00.0` in Groonga. Because Groonga uses the default value for NULL.
`index_column_diff` uses `00:00:00.0` for NULL in the expected posting lists. It causes false positive.

If we use `00:00:00.0` instead of ignoring for NULL, we can avoid the false positive.
It also enables that we can search NULL column value record by `00:00:00.0`.
In general, it'll be useful but this is an incompatible change. 
But it'll be acceptable because NULL behavior is undefined by definition.


otegami:
For reviewers, below are the results obtained when I checked the index values after updating NULL.
```sql
DROP TABLE IF EXISTS running_records;
CREATE TABLE running_records (
name VARCHAR(255),
average TIME(1) NULL,
KEY average_index(average)
) DEFAULT CHARSET=utf8mb4;
INSERT INTO running_records VALUES ("zero_condition", 0.0);
INSERT INTO running_records VALUES ("record failure", NULL);
SELECT mroonga_command("index_column_diff --table running_records#average_index --name index");
mroonga_command("index_column_diff --table running_records#average_index --name index")
[]
SELECT mroonga_command("select --table running_records#average_index --output_pretty yes");
mroonga_command("select --table running_records#average_index --output_pretty yes")
[
    [
      [
        1
      ],
      [
        [
          "_id",
          "UInt32"
        ],
        [
          "_key",
          "Time"
        ],
        [
          "index",
          "UInt32"
        ]
      ],
      [
        1,
        0.0,
        4
      ]
    ]
  ]
SELECT * FROM running_records WHERE average = 0;
name	average
zero_condition	00:00:00.0
record failure	00:00:00.0
```

mysql-test/mroonga/storage/column/time/fractional_seconds/t/null.test
```diff
@@ -0,0 +1,42 @@
...
+DROP TABLE IF EXISTS running_records;
+--enable_warnings
+
+CREATE TABLE running_records (
+  name VARCHAR(255),
+  average TIME(1) NULL,
+  KEY average_index(average)
+) DEFAULT CHARSET=utf8mb4;
+
+INSERT INTO running_records VALUES ("zero condition", 0.0);
```
kou, line 32:
```suggestion
INSERT INTO running_records VALUES ("zero condition", '00:00:00.0');
```

mysql-test/mroonga/storage/column/time/fractional_seconds/t/null.test
```diff
@@ -0,0 +1,42 @@
...
+INSERT INTO running_records VALUES ("record failure", NULL);
+
+SELECT mroonga_command("index_column_diff --table running_records#average_index --name index");
+
+SELECT * FROM running_records WHERE average = 0.0;
```
kou, line 37:
```suggestion
SELECT * FROM running_records WHERE average = '00:00:00.0';
```

mysql-test/mroonga/storage/column/time/fractional_seconds/t/null.test
```diff
@@ -0,0 +1,42 @@
...
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+--source ../../../../../include/mroonga/have_mroonga.inc
```
kou, line 19:
```suggestion
--source ../../../../../include/mroonga/have_fractional_seconds.inc
--source ../../../../../include/mroonga/have_mroonga.inc
```

kou: Hmm. It seems that all supported MySQL/MariaDB supports fractional seconds.
So we don't need to use `time/fractional_seconds/`.

kou: Hmm. `time/{fractional_seconds/,}t/with_index.test` is better than `time/with_index/t/{with,without}_fractional_seconds.test`. Let's use keep this path.

```console
$ find mysql-test/mroonga/storage/column/time
mysql-test/mroonga/storage/column/time
mysql-test/mroonga/storage/column/time/t
mysql-test/mroonga/storage/column/time/t/with_index.test
mysql-test/mroonga/storage/column/time/r
mysql-test/mroonga/storage/column/time/r/with_index.result
mysql-test/mroonga/storage/column/time/fractional_seconds
mysql-test/mroonga/storage/column/time/fractional_seconds/t
mysql-test/mroonga/storage/column/time/fractional_seconds/t/with_index.test
mysql-test/mroonga/storage/column/time/fractional_seconds/r
mysql-test/mroonga/storage/column/time/fractional_seconds/r/with_index.result
```

mysql-test/mroonga/storage/column/time/fractional_seconds/t/null.test
otegami: fix: [b3787d0](https://github.com/mroonga/mroonga/pull/791/commits/b3787d01a59f2766ff90b6a4df907fdf89a8c978)

mysql-test/mroonga/storage/column/time/t/null.test
```diff
@@ -0,0 +1,42 @@
...
+--source ../../../../include/mroonga/have_mroonga.inc
+--source ../../../../include/mroonga/load_mroonga_functions.inc
+
+--disable_warnings
+DROP TABLE IF EXISTS running_records;
+--enable_warnings
+
+CREATE TABLE running_records (
+  name VARCHAR(255),
+  average TIME(1) NULL,
```
kou, line 28: fractional secondsを使うならもとのように`time/fractional_seconds/`に置いておいてもらえますか？

mysql-test/mroonga/storage/column/time/fractional_seconds/t/null.test
otegami: fix: [8fc34c3](https://github.com/mroonga/mroonga/pull/791/commits/8fc34c371426e4f2fee916f4076d5a1b0d75ba86)
Ahh Isee. Could we remove the `MRN_HAVE_MYSQL_TYPE_TIME2` flag from the source code now?
If so,  I will remove it in another PR.
```cpp
#ifdef MRN_HAVE_MYSQL_TYPE_TIME2
  case MYSQL_TYPE_TIME2:
    error = generic_store_bulk_time2(field, buf);
    break;
#endif
```

kou: Yes.

mysql-test/mroonga/storage/column/time/t/null.test
otegami: `ha_mroonga::generic_store_bulk_time2()`に入る場合をテストしたかったので、`time/fractional_seconds/`配下において、`fractional seconds`を利用する形で行こうと思います。

PR #793, opened 2024-10-08
datetime fraction seconds: update index for NULL with `1970-01-01 00:00:00.0` to support index_column_diff

otegami:
GitHub: GH-789

The current implementation ignores NULL. It means that we don't update index for NULL.

`DATETIME` with fraction seconds' NULL is processed as `1970-01-01 00:00:00.0` in Groonga. Because Groonga uses the default value for NULL. `index_column_diff` uses `1970-01-01 00:00:00.0` for NULL in the expected posting lists. It causes false positive.

If we use `1970-01-01 00:00:00.0` instead of ignoring for NULL, we can avoid the false positive. It also enables that we can search NULL column value record by `1970-01-01 00:00:00.0`. In general, it'll be useful but this is an incompatible change. But it'll be acceptable because NULL behavior is undefined by definition.


otegami:
I will check the CI fails on windows.

PR #795, opened 2024-10-08
{float,double}: update index for `NULL` with `0.0` to support `index_column_diff`

otegami:
GitHub: GH-789

The current implementation ignores `NULL`. It means that we don't update index for `NULL`.

The `float` and `double` with `NULL` is processed as `0.0` in Groonga. Because Groonga uses the default value for `NULL`.
`index_column_diff` uses `0.0` for `NULL` in the expected posting lists. It causes false positive.

If we use `0.0` instead of ignoring for `NULL`, we can avoid the false positive.
It also enables that we can search `NULL` column value record by `0.0`.
In general, it'll be useful but this is an incompatible change.
But it'll be acceptable because `NULL` behavior is undefined by definition.

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -7349,7 +7349,8 @@ int ha_mroonga::storage_write_row(mrn_write_row_buf_t buf)
                              (field->real_type() != MYSQL_TYPE_LONG) &&
                              (field->real_type() != MYSQL_TYPE_TIME2) &&
                              (field->real_type() != MYSQL_TYPE_DATETIME2) &&
-                             (field->real_type() != MYSQL_TYPE_FLOAT)))
+                             (field->real_type() != MYSQL_TYPE_FLOAT) &&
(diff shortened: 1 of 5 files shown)


otegami:
I will rebase main.

ha_mroonga.cpp
```diff
@@ -12245,7 +12247,10 @@ int ha_mroonga::generic_store_bulk_float(Field* field, grn_obj* buf)
 {
   MRN_DBUG_ENTER_METHOD();
   int error = 0;
-  double value = field->val_real();
+  double value = 0;
```
kou, line 12250:
```suggestion
  double value = 0.0;
```

`0`でも動くんだけど、`double`で宣言しつつ初期化しているので`double`なリテラルを使ったほうがベターです。

otegami: たしかにです。ありがとうございます！

PR #796, opened 2024-10-09
timestamp fraction seconds: update index for `NULL` with `0000-00-00 00:00:00.0` to support `index_column_diff`

otegami:
GitHub: GH-789

The current implementation ignores `NULL`. It means that we don't update index for `NULL`. But it causes `index_column_diff` false positive.
`TIMESTAMP` with fraction seconds' `NULL` is processed as `0000-00-00 00:00:00.0` in Groonga. Because Groonga uses the default value for `NULL`. `index_column_diff` uses `0000-00-00 00:00:00.0` for `NULL` in the expected posting lists. It causes false positive.

If we use `0000-00-00 00:00:00.0` instead of ignoring for `NULL`, we can avoid the false positive. It also enables that we can search `NULL` column value record by `0000-00-00 00:00:00.0`. In general, it'll be useful but this is an incompatible change. But it'll be acceptable because `NULL` behavior is undefined by definition.

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -7350,7 +7350,8 @@ int ha_mroonga::storage_write_row(mrn_write_row_buf_t buf)
                              (field->real_type() != MYSQL_TYPE_TIME2) &&
                              (field->real_type() != MYSQL_TYPE_DATETIME2) &&
                              (field->real_type() != MYSQL_TYPE_FLOAT) &&
-                             (field->real_type() != MYSQL_TYPE_DOUBLE)))
+                             (field->real_type() != MYSQL_TYPE_DOUBLE) &&
+                             (field->real_type() != MYSQL_TYPE_TIMESTAMP2)))
       continue;
 
 #ifdef MRN_SUPPORT_GENERATED_COLUMNS
@@ -12390,10 +12391,13 @@ int ha_mroonga::generic_store_bulk_timestamp2(Field* field, grn_obj* buf)
 {
   MRN_DBUG_ENTER_METHOD();
   int error = 0;
-  Field_timestampf* timestamp_field = static_cast<Field_timestampf*>(field);
+  int64_t grn_time = 0;
+  Field_timestampf *timestamp_field = static_cast<Field_timestampf *>(field);
   mrn::TimestampFieldValueConverter<Field_timestampf> converter(
     timestamp_field);
-  int64_t grn_time = converter.convert();
+  if (!field->is_null()) {
+    grn_time = converter.convert();
+  }
   grn_obj_reinit(ctx, buf, GRN_DB_TIME, 0);
   GRN_TIME_SET(ctx, buf, grn_time);
   DBUG_RETURN(error);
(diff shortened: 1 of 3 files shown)


ha_mroonga.cpp
```diff
@@ -12385,10 +12386,13 @@ int ha_mroonga::generic_store_bulk_timestamp2(Field* field, grn_obj* buf)
 {
   MRN_DBUG_ENTER_METHOD();
   int error = 0;
-  Field_timestampf* timestamp_field = static_cast<Field_timestampf*>(field);
+  int64_t grn_time = 0;
+  Field_timestampf *timestamp_field = static_cast<Field_timestampf *>(field);
   mrn::TimestampFieldValueConverter<Field_timestampf> converter(
     timestamp_field);
```
kou, line 12392: Could you move this to the below `if` block to reduce scope?

mysql-test/mroonga/storage/column/timestamp/fractional_seconds/t/null.test
```diff
@@ -0,0 +1,43 @@
...
+
+CREATE TABLE diaries (
+  name VARCHAR(255),
+  created_at TIMESTAMP(1) NULL,
+  KEY created_at_index(created_at)
+) DEFAULT CHARSET=utf8mb4;
+
+INSERT INTO diaries VALUES ("zero day", "0000-00-00 00:00:00.0");
+INSERT INTO diaries VALUES ("null day", NULL);
+INSERT INTO diaries VALUES ("Mroonga release day", "2010-8-19 00:00:00.0");
```
kou, line 34: "time" is better than "day" because type is `TIMESTAMP`.

otegami: fix: [62194e5](https://github.com/mroonga/mroonga/pull/796/commits/62194e52c96973ef7c1be0886c8a85271387eb66) Sure!

mysql-test/mroonga/storage/column/timestamp/fractional_seconds/t/null.test
```diff
@@ -0,0 +1,43 @@
...
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+
+--source ../../../../../include/mroonga/have_mroonga.inc
```
kou, line 19:
```suggestion
# MySQL enables STRICT_ALL_TABLES by default and it rejects "0000-00-00 00:00:00.0".
--source ../../../../../include/mroonga/have_mariadb.inc
--source ../../../../../include/mroonga/have_mroonga.inc
```

otegami: Thanks you so much!

PR #799, opened 2024-10-09
enum: update index for `NULL` with `0` to support `index_column_diff`

otegami:
GitHub: GH-789

The current implementation ignores `NULL`. It means that we don't update index for `NULL`. But it causes `index_column_diff` false positive. `ENUM` with `NULL` is processed as `0` in Groonga. Because Groonga uses the default value for `NULL`. `ENUM` value is a string in user level but it's integer in MySQL/MariaDB internal. `NULL` `ENUM` value is `NULL` in MySQL/MariaDB internal and invalid `ENUM` value (`''`) is `0` in MySQL/MariaDB internal. We use `0` not `NULL` for Groonga because Groonga doesn't support `NULL`. The `index_column_diff` uses `0` for `NULL` in the expected posting lists. It causes false positive.

If we use `0` instead of ignoring for `NULL`, we can avoid the false positive. It also enables that we can search `NULL` column value record by `''` (or `0`).

In general, it'll be useful but this is an incompatible change. But it'll be acceptable because `NULL` behavior is undefined by definition.

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -7351,7 +7351,8 @@ int ha_mroonga::storage_write_row(mrn_write_row_buf_t buf)
                              (field->real_type() != MYSQL_TYPE_DATETIME2) &&
                              (field->real_type() != MYSQL_TYPE_FLOAT) &&
                              (field->real_type() != MYSQL_TYPE_DOUBLE) &&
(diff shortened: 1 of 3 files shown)


kou:
SQLでの文字列リテラルは`"..."`じゃなくて`'...'`の方がベターです。
SQL標準では`'...'`だからです。

PR #807, opened 2024-10-11
date: update index for `NULL` with `1970-01-01` to support `index_column_diff`

otegami:
GitHub: GH-789

The current implementation ignores `1970-01-01`. It means that we don't update index for `NULL`. But it causes `index_column_diff` false positive.

`DATE` with `NULL` is processed as `0` (`1970-01-01` in MySQL/MariaDB) in Groonga. Because Groonga uses the default value for `NULL`. `index_column_diff` uses `0` for `NULL` in the expected posting lists. It causes false positive.

If we use `0`( `1970-01-01` in MySQL/MariaDB) instead of ignoring for `NULL`, we can avoid the false positive. It also enables that we can search `NULL` column value record by `1970-01-01`. In general, it'll be useful but this is an incompatible change. But it'll be acceptable because NULL behavior is undefined by definition.

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -12496,12 +12496,12 @@ int ha_mroonga::generic_store_bulk_new_date(Field* field, grn_obj* buf)
   MRN_DBUG_ENTER_METHOD();
   int error = 0;
   bool truncated = false;
-  Field_newdate* newdate_field = (Field_newdate*)field;
-  mrn::TimeConverter time_converter;
   long long int time = 0;
   if (!field->is_null()) {
+    Field_newdate* newdate_field = (Field_newdate*)field;
     MYSQL_TIME mysql_date;
     MRN_FIELD_GET_TIME(newdate_field, &mysql_date, current_thd);
+    mrn::TimeConverter time_converter;
     time = time_converter.mysql_time_to_grn_time(&mysql_date, &truncated);
   }
   if (truncated) {
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -12498,7 +12498,7 @@ int ha_mroonga::generic_store_bulk_new_date(Field* field, grn_obj* buf)
   bool truncated = false;
   long long int time = 0;
   if (!field->is_null()) {
-    Field_newdate* newdate_field = (Field_newdate*)field;
+    auto newdate_field = static_cast<Field_newdate*>(field);
     MYSQL_TIME mysql_date;
     MRN_FIELD_GET_TIME(newdate_field, &mysql_date, current_thd);
     mrn::TimeConverter time_converter;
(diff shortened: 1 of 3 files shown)


ha_mroonga.cpp
```diff
@@ -12496,11 +12497,13 @@ int ha_mroonga::generic_store_bulk_new_date(Field* field, grn_obj* buf)
   int error = 0;
   bool truncated = false;
   Field_newdate* newdate_field = (Field_newdate*)field;
-  MYSQL_TIME mysql_date;
-  MRN_FIELD_GET_TIME(newdate_field, &mysql_date, current_thd);
   mrn::TimeConverter time_converter;
```
kou, line 12500: これもブロックに入れられます？

mysql-test/mroonga/storage/column/date/t/null.test
```diff
@@ -0,0 +1,43 @@
...
+
+CREATE TABLE diaries (
+  name VARCHAR(255),
+  created_on DATE NULL,
+  KEY created_on_index(created_on)
+) DEFAULT CHARSET=utf8mb4;
+
+INSERT INTO diaries VALUES ('epoch date', '1970-01-01');
+INSERT INTO diaries VALUES ('null date', NULL);
+INSERT INTO diaries VALUES ('Mroonga release day', '2010-8-19');
```
kou, line 34: date?

ha_mroonga.cpp
otegami: fix: [93e22b7](https://github.com/mroonga/mroonga/pull/807/commits/93e22b7867ea71bbb79e910cc3b8abd11f893d0e) 合わせて、`newdate_field`も入れられそうだったので入れておきました :pray:

mysql-test/mroonga/storage/column/date/t/null.test
otegami: fix: [5102283](https://github.com/mroonga/mroonga/pull/807/commits/51022837796d153dc7f1fa975ec1b0086e4d2531) (epoch timeは、直せたのにmm)

ha_mroonga.cpp
```diff
@@ -12495,12 +12496,14 @@ int ha_mroonga::generic_store_bulk_new_date(Field* field, grn_obj* buf)
...
   bool truncated = false;
-  Field_newdate* newdate_field = (Field_newdate*)field;
-  MYSQL_TIME mysql_date;
-  MRN_FIELD_GET_TIME(newdate_field, &mysql_date, current_thd);
-  mrn::TimeConverter time_converter;
-  long long int time =
-    time_converter.mysql_time_to_grn_time(&mysql_date, &truncated);
+  long long int time = 0;
+  if (!field->is_null()) {
+    Field_newdate* newdate_field = (Field_newdate*)field;
```
kou, line 12501:
```suggestion
    auto newdate_field = static_cast<Field_newdate*>(field);
```

せっかくなので新しい書き方にしましょうか。

otegami: fix: [3e917e5](https://github.com/mroonga/mroonga/pull/807/commits/3e917e528aba226d8f1cc55f07242ea6a679329f) はい！

PR #832, opened 2024-11-15
docs: update MariaDB version from 14.4.3 to 14.4.4

otegami:


otegami:
I want to think of the way we can automatically update the version in documentation.

kou:
We don't want to maintain MariaDB version here.

FYI: Sphinx has the substitution feature:
* https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#substitutions
* https://github.com/mroonga/mroonga/blob/main/doc/source/install/windows.rst#install
* https://github.com/mroonga/mroonga/blob/a8a6009ce0d3ab6db64846905029258861140160/doc/source/conf.py#L133

But it doesn't work in pre-formatted text part...

doc/source/developer/debug.md
```diff
@@ -41,17 +41,17 @@ The procedure from download to build is the following:
 ```console
 $ mkdir -p ~/work/
 $ cd ~/work/
-$ wget https://downloads.mariadb.org/rest-api/mariadb/11.4.3/mariadb-11.4.3.tar.gz
-$ tar xf mariadb-11.4.3.tar.gz
-$ rm -rf mariadb-11.4.3/storage/mroonga
+$ wget https://downloads.mariadb.org/rest-api/mariadb/11.4.4/mariadb-11.4.4.tar.gz
```
kou, line 44: We can use https://downloads.mariadb.org/rest-api/mariadb/ to detect the latest MariaDB.

otegami: Thanks. I will think how to handle the response from https://downloads.mariadb.org/rest-api/mariadb/.

Issue #837, opened 2025-01-01
ci: add install jobs on AlmaLinux


kou:
### What is your problem?

We are using the latest MariaDB/MySQL/Percona Server automatically for AlmaLinux packages. So our existing CI can't detect that our released packages for AlmaLinux are outdated.

We should add AlmaLinux jobs to https://github.com/mroonga/mroonga/blob/main/.github/workflows/install.yml to detect outdated AlmaLinux packages.

komainu8:
I added AlmaLinux jobs to https://github.com/mroonga/mroonga/blob/main/.github/workflows/install.yml to detect outdated AlmaLinux packages.

PR #843, opened 2025-01-14
ci: add install test for AlmaLinux 9 with MariaDB 10.6

komainu8:
Related: GitHub: GH-837
This adds support for only AlmaLinux 9 and Mariadb 10.6.


komainu8:
https://github.com/komainu8/mroonga/actions/runs/12760026458/job/35564779637

PR #846, opened 2025-01-14
ci: add install test for AlmaLinux 8 with MariaDB 11.4

komainu8:
Related: GitHub: GH-837
This adds support for only AlmaLinux 8 and Mariadb 11.4.


komainu8:
https://github.com/komainu8/mroonga/actions/runs/12761714128/job/35569423355

PR #849, opened 2025-01-15
ci: add install test for AlmaLinux 8 with MySQL Community Server 8.0

komainu8:
Related: GitHub: GH-837
This adds support for only AlmaLinux 8 and MySQL Community Server 8.0.


komainu8:
https://github.com/komainu8/mroonga/actions/runs/12779271845/job/35623596516

PR #850, opened 2025-01-15
ci: add install test for AlmaLinux 9 with MySQL Community Server 8.0

komainu8:
Related: GitHub: GH-837
This adds support for only AlmaLinux 9 and MySQL Community Server 8.0.


komainu8:
https://github.com/komainu8/mroonga/actions/runs/12782010072/job/35630837621

PR #862, opened 2025-01-27
doc: Add the `--quiet` and `--fail-on-warning` options to `sphinx-build`

abetomo:
* `--quiet`
  * It is easier to see output related to errors
* `--fail-on-warning`
  * This option causes the build to fail on warnings
  * It makes it easier to notice mistakes
* Added `red-datasets` to Gemfile because of warning output at build

These are the same as the settings for building documents in groonga/groonga.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -67,9 +67,10 @@ function(mrn_sphinx SOURCE_DIR LOCALE SOURCES HTML_FILES)
       COMMAND
         ${CMAKE_COMMAND} -E env DOCUMENT_VERSION=${MRN_VERSION_FULL}
         DOCUMENT_VERSION_FULL=${MRN_VERSION_FULL} LOCALE=${LOCALE}
-	PACKAGE_MROONGA_VERSION=${MRN_VERSION_FULL}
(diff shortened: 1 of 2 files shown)


doc/CMakeLists.txt
```diff
@@ -67,9 +67,10 @@ function(mrn_sphinx SOURCE_DIR LOCALE SOURCES HTML_FILES)
       COMMAND
         ${CMAKE_COMMAND} -E env DOCUMENT_VERSION=${MRN_VERSION_FULL}
         DOCUMENT_VERSION_FULL=${MRN_VERSION_FULL} LOCALE=${LOCALE}
-	PACKAGE_MROONGA_VERSION=${MRN_VERSION_FULL}
-        ${SPHINX_BUILD} -j auto -D language=${LOCALE} -b ${BUILDER} -d
-        ${LOCALE}/doctrees/${BUILDER} ${SOURCE_DIR} ${LOCALE}/${BUILDER}
+        PACKAGE_MROONGA_VERSION=${MRN_VERSION_FULL}
```
abetomo, line 70: Only this line was a hard tab, so it was changed to a soft tab.

PR #878, opened 2025-02-06
packages yum: change mysql client name to mariadb on MariaDB

komainu8:
MariaDB 11.4.5 stopped providing `mysql` command. We need to use `mariadb` not `mysql`.

MariaDB 10.X still provides `mysql` command but we can use `mariadb` with MariaDB 10.X.

### DIFF
--- a/packages/yum/test.sh
+++ b/packages/yum/test.sh
@@ -48,7 +48,7 @@ case ${package} in
     old_package=${package}
     mysql_package_prefix=MariaDB
     service_name=mariadb
-    mysql_client=mariadb
+    mysql_command=mariadb
     ha_mroonga_so=ha_mroonga_official.so
     test_package_name=MariaDB-test
     baseurl=https://yum.mariadb.org/${mysql_version}/rhel/${major_version}/x86_64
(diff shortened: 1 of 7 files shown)


packages/yum/test.sh
```diff
@@ -278,6 +282,11 @@ echo "::group::Upgrade"
 if [ "${triggered_ref_type}" = "tag" ]; then
   echo "Skip on release because external dependency updates of old package " \
        "cause test failures."
+elif [ "${mysql_version}" = "11.4" ]; then
+  echo "TODO: We can remove this after we release Mroonga 15.00." \
+       "MariaDB client name is changed to mariadb from mysql since MariaDB 11.4.5." \
+       "Mroonga 14.14 or eailr use mysql command. So, upgrade test is always fail currently." \
+       "We disable upgrade test temporary."
```
kou, line 289:
```suggestion
  echo "TODO: We can remove this after we release Mroonga 15.00."
  echo "MariaDB client name is changed to mariadb from mysql since MariaDB 11.4.5."
  echo "Mroonga 14.14 or eailr use mysql command. So, upgrade test is always fail currently."
  echo "We disable upgrade test temporary."
```

PR #902, opened 2025-03-28
mroonga_* udf correct ptr types for is_null/error

grooverdan:
Shows up in mroonga UDF tests under clang with UBSAN:

UndefinedBehaviorSanitizer: function-type-mismatch

### DIFF
--- a/udf/mrn_udf_command.cpp
+++ b/udf/mrn_udf_command.cpp
@@ -268,7 +268,7 @@ static void mroonga_command_escape_value(grn_ctx *ctx,
 }
 
 MRN_API char *mroonga_command(UDF_INIT *init, UDF_ARGS *args, char *result,
-                              unsigned long *length, char *is_null, char *error)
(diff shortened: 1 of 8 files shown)


kou:
Oh, I didn't notice it.

It seems that MariaDB uses `uchar` not `unsigned char`: https://github.com/MariaDB/server/blob/22efc2c784e1b7199fb5804e6330168277ea7dce/sql/sql_udf.h#L60

How about using `uchar` here too?

BTW, it seems that `sql/udf_example.c` in MariaDB also uses `char` not `uchar`: https://github.com/MariaDB/server/blob/22efc2c784e1b7199fb5804e6330168277ea7dce/sql/udf_example.c#L159

grooverdan:
yep, you're right on uchar. Thanks.

example fix was in progress - https://github.com/MariaDB/server/pull/3906 but can update.

kou:
Thanks.

PR #906, opened 2025-04-03
[WIP] Add support utf8mb4_uca1400_ai_ci

komainu8:
TODO: Add tests.

### DIFF
--- a/lib/mrn_field_normalizer.cpp
+++ b/lib/mrn_field_normalizer.cpp
@@ -1,7 +1,7 @@
 /* -*- c-basic-offset: 2 -*- */
 /*
   Copyright(C) 2013-2018  Kouhei Sutou <kou@clear-code.com>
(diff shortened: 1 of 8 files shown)


otegami:
Thanks!!

Each collation's suffix means as follows.
- "ai"（accent insensitive）
- "as"（accent sensitive）
- "ci"（case insensitive）
- "cs"（case sensitive）

ref: https://github.com/groonga/groonga-normalizer-mysql
ref: https://mariadb.com/kb/en/supported-character-sets-and-collations/
ref: https://www.unicode.org/reports/tr10/#Multi_Level_Comparison

otegami:
We will handle it in https://github.com/mroonga/mroonga/pull/908 instead of this PR.

otegami:
@komainu8 
We've done this change in GH-908. Could you close this PR?

PR #925, opened 2025-05-12
ci: windows: Sleep before getting the patch

abetomo:
This is a temporary workaround.
If we access GitHub immediately, we get the following error:

```
Whoa there!
You have exceeded a secondary rate limit.
...
```

Sleep for 65 seconds to avoid.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -178,6 +178,10 @@ jobs:
           }
           Remove-Item extra\groonga\test -Recurse -Force
           cd extra\groonga\vendor\mruby-source
+          # Temporary workaround.
+          # If I access github immediately, an error occurs:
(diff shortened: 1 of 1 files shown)


abetomo:
It seems to have started failing a week ago.
https://github.com/mroonga/mroonga/actions/runs/14827081351/job/41621757650#step:15:70

komainu8 (approved):

otegami:
Could we use `-MaximumRetryCount` and `RetryIntervalSec`?
- https://learn.microsoft.com/en-us/powershell/module/Microsoft.PowerShell.Utility/Invoke-WebRequest?view=powershell-7.4#-maximumretrycount

abetomo:
Thanks for the info.

PR #927, opened 2025-05-12
ci: windows: Set RetryIntervalSec for patch fetching to 65 seconds

abetomo:
If we access GitHub immediately, we get the following error:

```
Whoa there!
You have exceeded a secondary rate limit.
...
```

Wait for 65 seconds to avoid.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -179,6 +179,8 @@ jobs:
           Remove-Item extra\groonga\test -Recurse -Force
           cd extra\groonga\vendor\mruby-source
           Invoke-WebRequest `
+            -MaximumRetryCount 3 `
+            -RetryIntervalSec 65 `
             -Uri "https://github.com/mruby/mruby/pull/6279.patch" `
             -OutFile 6279.patch
           ridk exec patch -p1 -i 6279.patch


abetomo:
See also: #925

.github/workflows/windows.yml
```diff
@@ -179,6 +179,8 @@ jobs:
           Remove-Item extra\groonga\test -Recurse -Force
           cd extra\groonga\vendor\mruby-source
           Invoke-WebRequest `
+            -MaximumRetryCount 3 `
+            -RetryIntervalSec 65 `
```
otegami, line 183: We might know when the rate limit will be reset from response header's `x-ratelimit-reset`.
If the value is around 65, it will be better. Or how about using just 60 seconds because rate limit seems to be reset evert 60 seconds?   
https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#checking-the-status-of-your-rate-limit

abetomo: `x-ratelimit-reset` header was not included in the response.
Maybe because it is not an API.

I'm guessing that the limit is 60 seconds, but I set it a little more at 65 seconds because it might be denied if it is just right.

otegami: Thank you for checking it. It sounds nice to me!

otegami (approved):
LGTM 👍🏾

komainu8 (approved):

PR #931, opened 2025-05-20
ci: run release jobs as much as possible

kou:
It's for avoiding to detect a problem on tag push.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,8 +74,7 @@ jobs:
           if [ "${GITHUB_REF_TYPE}" = "tag" ]; then
             version=${GITHUB_REF_NAME#v}
           else
-            version=$(head -n1 release-note.md | \
-                        grep -E -o '[0-9]+\.[0-9]+')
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -56,26 +65,38 @@ jobs:
...
+            tee release-note.md
+          echo "" | tee -a release-note.md
+          echo "### Translations" | tee -a release-note.md
+          echo "" | tee -a release-note.md
+
+          if [ "${GITHUB_REF_TYPE}" = "tag" ]; then
+            version=${GITHUB_REF_NAME#v}
+          else
+            version=$(head -n1 release-note.md | \
+                        grep -E -o '[0-9]+\.[0-9]+\.[0-9]+')
```
otegami, line 78:
```suggestion
                        grep -E -o '[0-9]+\.[0-9]+')
```

otegami (commented):
Thank you so much!

.github/workflows/release.yml
kou: Good catch!

otegami (approved):

PR #932, opened 2025-05-23
Update name of source base directory 

komainu8:
Currently,  direcotry name is "mariadb-10.6-10.6.22" instead of "mariadb-10.6.22" after expand source archive as below.

---

$ tar -zxvf mariadb-10.6_10.6.22.orig.tar.gz
$ la
drwxr-xr-x 34 aaa   aaa   4.0K  5月  7 13:27  mariadb-10.6-10.6.22/
-rw-r--r--  1 aaa   aaa    97M  5月 20 20:36  mariadb-10.6_10.6.22.orig.tar.gz


abetomo:
The following will succeed whatever the directory name of the archive is, is it not better?

```
rm -rf $${source_base} && mkdir -p $${source_base}
(cd $${source_base} &&
  tar xf ../$${archive} --strip-components 1 &&
  tar xf ../$${debian})
```

komainu8:
Thank you for your comments.
I've fixed them.

otegami (approved):
LGTM if CI passes.

PR #948, opened 2025-07-14
ci: mariadb: Run Configure MariaDB in compile-only-minimal as well

abetomo:
We are trying to submodule Mroonga in the MariaDB server. 
But we get an error, so we test if it builds as expected.

`Configure MariaDB` with `id: amd64-compile-only-minimal` fails because it is a commit that changes CI in preparation for the upcoming fix.

### DIFF
--- a/.github/workflows/mariadb.yml
+++ b/.github/workflows/mariadb.yml
@@ -26,9 +26,90 @@ concurrency:
   cancel-in-progress: true
 jobs:
   main:
-    name: main
+    name: ${{ matrix.id }}
     timeout-minutes: 30
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # To reduce build time, we build only the MariaDB core and Mroonga.
+          # The following options disable all other storage engines, module-type
+          # plugins, and some core options to reduce compile time and external
+          # dependencies.
+          #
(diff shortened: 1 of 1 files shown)


abetomo:
Related:

* https://github.com/MariaDB/server/pull/4185
* https://buildbot.mariadb.org/#/builders/906/builds/23

.github/workflows/mariadb.yml
```diff
@@ -68,80 +149,25 @@ jobs:
...
           restore-keys: mariadb-linux-ccache-
       - name: Configure MariaDB
-        # To reduce build time, we build only the MariaDB core and Mroonga.
-        # The following options disable all other storage engines, module-type
-        # plugins, and some core options to reduce compile time and external
-        # dependencies.
-        #
-        # To list available plugins, you can run:
-        #   cmake -L $MARIADB_SOURCE_DIR | grep PLUGIN_
+        continue-on-error: ${{ matrix.id != 'amd64-compile-only-core' && false || true }}
```
otegami, line 152: It looks like `continue-on-error` always evaluates to true regardless of the matrix value. Is my understanding correct? If so, we want to make it false using `amd64-compile-only-core`.

otegami (approved):

komainu8 (approved):

PR #957, opened 2025-07-31
release yum: use URI that returns newer SRPM version

komainu8:
The CDN caches both “…/SRPMS” and "…/SRPMS/".
Also, both of caches may return old version.

We will try that we get the latest version from `dnf` command.

---

I have already tried the following approaches, but these are not working out.

* Get the latest version from tags on GitHub(mysql/mysql-server).
* Get the latest version from MySQL's release note.

These approaches can not get MySQL release number.
(MySQL release number is 1 of mysql-community-8.0.43-1.el9.src.rpm.)
Probably, this number can only get from MySQL yum repository.
(We don't get this number from release note and Github tags...)

### DIFF
--- a/packages/mroonga-package-task.rb
+++ b/packages/mroonga-package-task.rb
@@ -238,6 +238,10 @@ class MroongaPackageTask < PackagesGroongaOrgPackageTask
   def latest_target_version(series, trailing_slash, minimal_or_not = nil)
     srpms_url =
       "https://repo.mysql.com/yum/mysql-#{series}-community/el/9/SRPMS"
+    if minimal_or_not == :minimal
+      srpms_url =
+        "https://repo.mysql.com/yum/mysql-#{series}-community/docker/el/9/SRPMS"
+    end
     pattern = "SRPMS\/mysql-community-"
     if trailing_slash == :with_trailing_slash
       srpms_url << "/"
--- a/packages/mroonga-package-task.rb
+++ b/packages/mroonga-package-task.rb
@@ -281,19 +281,6 @@ class MroongaPackageTask < PackagesGroongaOrgPackageTask
       latest_target_version(series, :with_trailing_slash, :minimal),
       latest_target_version(series, :without_trailing_slash, :minimal)
     ].max
-
-    srpms_url =
-      "https://repo.mysql.com/yum/mysql-#{series}-community/docker/el/9/SRPMS"
-    index_html = URI.open(srpms_url) do |response|
-      response.read
-    end
-    latest_target_srpm =
-      index_html.
-        scan(/href="(.+?)"/i).
-        flatten.
-        grep(/\ASRPMS\/mysql-community-minimal-/).
-        last
-    latest_target_srpm[/\ASRPMS\/mysql-community-minimal-(\d+\.\d+\.\d+-\d+)/, 1]
   end
 
   def mysql_community_minimal_rpm_version
--- a/packages/mroonga-package-task.rb
+++ b/packages/mroonga-package-task.rb
@@ -235,7 +235,7 @@ class MroongaPackageTask < PackagesGroongaOrgPackageTask
     [names.join("-"), series]
   end
 
-  def latest_target_version(series, trailing_slash: false, minimal: false)
+  def target_version(series, trailing_slash: false, minimal: false)
     path = minimal ? "docker/el/9/SRPMS" : "el/9/SRPMS"
     path += "/" if trailing_slash
     srpms_url =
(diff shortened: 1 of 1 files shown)


packages/mroonga-package-task.rb
```diff
@@ -235,23 +235,42 @@ def split_mysql_package
     [names.join("-"), series]
   end
 
-  def detect_mysql_community_rpm_version
-    series = split_mysql_package[1]
-    # Use the URL without a trailing slash because the CDN caches both “…/SRPMS”
-    # and "…/SRPMS/", but the cache for the "/SRPMS/" path hasn't been expired
-    # yet. Using the no‑slash URL ensures we get the latest listing.
+  def latest_target_version(series, trailing_slash, minimal_or_not = nil)
```
abetomo, line 238:
```suggestion
  def latest_target_version(series, trailing_slash: false, minimal: false)
```

Call method: `latest_target_version(series, trailing_slash: true, minimal: true)` 

It seems better to use boolean in this way.

packages/mroonga-package-task.rb
```diff
@@ -235,23 +235,42 @@ def split_mysql_package
...
+        "https://repo.mysql.com/yum/mysql-#{series}-community/docker/el/9/SRPMS"
+    end
+    pattern = "SRPMS\/mysql-community-"
+    if trailing_slash
+      srpms_url << "/"
+      pattern = "mysql-community-"
+    end
+    if minimal
+      pattern << "minimal-"
+    end
```
otegami, line 252:
```suggestion
path = minimal ? "docker/el/9/SRPMS" : "el/9/SRPMS"
path += "/" if trailing_slash
srpms_url =
"https://repo.mysql.com/yum/mysql-#{series}-community/${path}"
pattern =  trailing_slash ? "SRPMS\/mysql-community-" : "mysql-community-"
pattern += "minimal-"
```

packages/mroonga-package-task.rb
```diff
@@ -235,23 +235,42 @@ def split_mysql_package
...
+  end
+
+  def detect_mysql_community_rpm_version
+    series = split_mysql_package[1]
+    # Use the URL returns newer version because the CDN caches both “…/SRPMS”
+    # and "…/SRPMS/", but one of caches may return old version.
+    [
+      latest_target_version(series, trailing_slash: true),
+      latest_target_version(series, trailing_slash: false)
+    ].max
```
otegami, line 273: I think using `Gem::Version` would be better when comparing the versions.

packages/mroonga-package-task.rb
```diff
@@ -235,23 +235,45 @@ def split_mysql_package
...
-    # Use the URL without a trailing slash because the CDN caches both “…/SRPMS”
-    # and "…/SRPMS/", but the cache for the "/SRPMS/" path hasn't been expired
-    # yet. Using the no‑slash URL ensures we get the latest listing.
+  def target_version(series, trailing_slash: false, minimal: false)
+    path = minimal ? "docker/el/9/SRPMS" : "el/9/SRPMS"
+    path += "/" if trailing_slash
     srpms_url =
-      "https://repo.mysql.com/yum/mysql-#{series}-community/el/9/SRPMS"
+      "https://repo.mysql.com/yum/mysql-#{series}-community/#{path}"
+    pattern =  trailing_slash ? "SRPMS\/mysql-community-" : "mysql-community-"
```
abetomo, line 243:
```suggestion
    pattern = trailing_slash ? "SRPMS\/mysql-community-" : "mysql-community-"
```

packages/mroonga-package-task.rb
```diff
@@ -235,23 +235,45 @@ def split_mysql_package
...
+
     index_html = URI.open(srpms_url) do |response|
       response.read
     end
     latest_target_srpm =
       index_html.
         scan(/href="(.+?)"/i).
         flatten.
-        grep(/\ASRPMS\/mysql-community-/).
+        grep(/\A#{pattern}/).
```
abetomo, line 253: https://github.com/mroonga/mroonga/pull/957/files#diff-8e242dc184bb0419df8870248a2bc87f09cc4471fecaf61624ac6bc797fdc92aR123

変数を埋め込むときは `Regexp.escape()` しておいたほうが良さげ？

packages/mroonga-package-task.rb
```diff
@@ -235,23 +235,45 @@ def split_mysql_package
...
         last
-    latest_target_srpm[/\ASRPMS\/mysql-community-(\d+\.\d+\.\d+-\d+)/, 1]
+    latest_target_srpm[/\A#{pattern}(\d+\.\d+\.\d+-\d+)/, 1]
```
otegami, line 255: How about returning the version using `Gem::Version.new`?

komainu8:
This problem has already resolved at #961.
So, I close this PR.

PR #959, opened 2025-08-06
release yum: set the latest MySQL Community version from MySQL's release note

komainu8:
TODO: Test on CI.

The CDN caches both “…/SRPMS” and "…/SRPMS/".
also, both of caches may return old version.

Therefore, we use version of MySQL's release note instead of URI of SRPMS.

---

We can also get the latest version from tags for mysql/mysql-server repository on GitHub.
However, I think this approach is not good. because of tags of mysql/mysql-server is many and we can get tags only every 100 tags via GItHub API.
And mysql/mysql-server doesn't have release page on GitHub. So, We can only use tags information to get the latest version.

The cause of tags of mysql/mysql-server is many is they also include mysql-cluster and  EOLed version (e.g. mysql-5.7.xx and mysql-5.6.xx)

### DIFF
--- a/packages/mroonga-package-task.rb
+++ b/packages/mroonga-package-task.rb
@@ -237,21 +237,19 @@ class MroongaPackageTask < PackagesGroongaOrgPackageTask
 
   def detect_mysql_community_rpm_version
     series = split_mysql_package[1]
-    # Use the URL without a trailing slash because the CDN caches both “…/SRPMS”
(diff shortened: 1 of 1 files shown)


komainu8:
We don't get MySQL release number when we get the latest version from MySQL's release note...

MySQL release number is `1` of  `mysql-community-8.0.43-1.el9.src.rpm`.
This number only get from MySQL yum repository.
(We don't get this number from release note and Github tags...)

So, this approach is not good too...

PR #960, opened 2025-08-07
package: Use the package information in the repository

abetomo:
The latest versions of MySQL and MariaDB are fetched from the web page.
But, the latest version cannot be fetched due to caching.

In such cases, fetch the latest version from the package in the repository.


kou:
How about using https://repo.mysql.com/yum/mysql-8.4-community/el/9/SRPMS/repodata/repomd.xml and `-primary.xml.gz` in the `repomd.xml`?

PR #961, opened 2025-08-11
package: Fetch the version from the package metadata

abetomo:
The latest versions of MySQL and MariaDB are fetched from the web page.
But, the latest version cannot be fetched due to caching.

Therefore, we decided to fetch the version from the package metadata.

### DIFF
--- a/packages/mroonga-package-task.rb
+++ b/packages/mroonga-package-task.rb
@@ -3,6 +3,8 @@ require "json"
 require "open-uri"
 require "pathname"
 require "pp"
+require "rexml/document"
+require "rexml/xpath"
 require "zlib"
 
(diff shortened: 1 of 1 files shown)


otegami:
I’d like to know whether this change works.
So if you’ve tried it locally, I’d like to see the success log.
If it’s hard to get it, that’s fine with me.

abetomo:
Before this fix, building RPM packages would fail.
For example: https://github.com/mroonga/mroonga/actions/runs/16896335540

This fix will succeed in building RPM.
https://github.com/mroonga/mroonga/actions/runs/16895794904?pr=961
(But MariaDB 10.5 fails because of EOL. #963)

otegami:
Thank you so much. I understood it!

otegami (approved):

Issue #996, opened 2025-10-27
test wrapper/fulltext/t/not_match_against: executing test got signal 6


komainu8:
### What happened?

```
mroonga/wrapper/fulltext.not_match_against w1 [ fail ]
        Test ended at 2025-09-11 12:15:07

CURRENT_TEST: mroonga/wrapper/fulltext.not_match_against
mysqltest: At line 41: query 'select * from t1 where c1 <= 4 and not match(c3) against("uu") order by c1' failed: <Unknown> (2013): Lost connection to server during query

The result from queries just before the failure was:
< snip >
2	10	ka ki ku ke ko
4	10	ka ki ku ke ko
6	20	ka ki ku ke ko
8	20	ka ki ku ke ko
select *,match(c3) against("dummy") from t1 where match(c3) against("dummy") order by c1;
c1	c2	c3	match(c3) against("dummy")
select * from t1 where not match(c3) against("dummy") order by c1;
c1	c2	c3
1	10	aa ii uu ee oo
2	10	ka ki ku ke ko
3	10	aa ii uu ee oo
4	10	ka ki ku ke ko
5	20	aa ii uu ee oo
6	20	ka ki ku ke ko
7	20	aa ii uu ee oo
8	20	ka ki ku ke ko
select * from t1 where c1 = 4 and not match(c3) against("uu") order by c1;
c1	c2	c3
4	10	ka ki ku ke ko
select * from t1 where c1 <= 4 and not match(c3) against("uu") order by c1;
```

### How to reproduce it

```
--source ../../../include/mroonga/have_mariadb.inc
--source ../../../include/mroonga/have_innodb.inc
--source ../../../include/mroonga/have_mroonga.inc

--disable_warnings
drop table if exists t1;
--enable_warnings

create table t1 (
  c1 int primary key
) COMMENT = 'engine "innodb"';

select * from t1 where c1 <= 4;

drop table t1;

--source ../../../include/mroonga/have_mroonga_deinit.inc
```

### Expected behavior

Test success.

### Environment

- OS: not related
- OS Version: not related
- DB: MariaDB
- DB Version: 11.8.3
- Mroonga Version: master
  - https://mroonga.org/docs/reference/server_variables.html#mroonga-version
- libgroonga Version: 15.1.7
  - https://mroonga.org/docs/reference/server_variables.html#mroonga-libgroonga-version

PR #998, opened 2025-10-29
wrapper: setting optimaizer costs when Mroonga create the new handler

komainu8:
Fix: GH-996.

MariaDB's handler initialize `optimizer_where_cost`(=`WHERE_COST`) to `0` by initialization list of contractor since MariaDB 11 or later.
See: https://github.com/MariaDB/server/commit/b66cdbd1eaeed7e96317a03a190c496fd062ec71

```diff
public:
  handler(handlerton *ht_arg, TABLE_SHARE *share_arg)
    :table_share(share_arg), table(0),
    estimation_rows_to_insert(0),
    lookup_handler(this),
-   ht(ht_arg), ref(0), lookup_buffer(NULL), end_range(NULL),
+   ht(ht_arg), costs(0), ref(0), lookup_buffer(NULL), end_range(NULL),
    implicit_emptied(0),
    mark_trx_read_write_done(0),
    check_table_binlog_row_based_done(0),
@@ -3364,7 +3379,6 @@ class handler :public Sql_alloc
    ref_length(sizeof(my_off_t)),
    ft_handler(0), inited(NONE), pre_inited(NONE),
    pushed_cond(0), next_insert_id(0), insert_id_for_cur_row(0),
    optimizer_cache_cost((100-DEFAULT_CACHE_HIT_RATIO)/100.0),
    tracker(NULL),
    pushed_idx_cond(NULL),
    pushed_idx_cond_keyno(MAX_KEY),
@@ -3378,12 +3392,19 @@ class handler :public Sql_alloc
    m_psi_numrows(0),
    m_psi_locker(NULL),
    row_logging(0), row_logging_init(0),
-   m_lock_type(F_UNLCK), ha_share(NULL)
+   m_lock_type(F_UNLCK), ha_share(NULL), optimizer_where_cost(0),
+   optimizer_scan_setup_cost(0)
  {
```

Mroonga use MariaDB's handler for passing process to InnoDB in the wrapper mode.
The `optimizer_where_cost` value of handler is set to `0` when Mroonga create new handler.

 In this way, The `optimizer_where_cost` value is set to non-expected value because of Mroonga update the `optimizer_where_cost` value of handler in the wrapper mode.

As a result, program is interrupted by assertion because of the validation of `optimizer_where_cost` is failure.

Mroonga should set current cost value after creating new handler by `set_optimizer_costs()`.

We passed the test(mroonga/wrapper/fulltext.not_match_against) as below by applying this modification.

---

`mroonga/wrapper/fulltext.not_match_against w2 [ pass ]     67`
https://github.com/komainu8/mroonga/actions/runs/18894502583/job/53929036016#step:15:424

---

In addition, this issue does not occur in storage mode because of storage mode does not create MariaDB's handler.

### DIFF
--- a/mrn_mysql_compat.h
+++ b/mrn_mysql_compat.h
@@ -896,6 +896,7 @@ using TABLE_LIST = Table_ref;
 // the warn_deprecated<>() definition for details.
 #  define MRN_WARN_DEPRECATED(thd, what, to)                                   \
     (warn_deprecated<999999>(thd, what, to))
+#  define MRN_HAVE_SET_OPTIMAIZER_COSTS
 #else
 #  define MRN_WARN_DEPRECATED(thd, what, to)                                   \
     (push_warning_printf(thd,                                                  \
@@ -904,10 +905,6 @@ using TABLE_LIST = Table_ref;
                          MRN_GET_ERR_MSG(ER_WARN_DEPRECATED_SYNTAX),           \
                          what,                                                 \
                          to))
-#endif
-
-#if defined(MRN_MARIADB_P) && (MYSQL_VERSION_ID >= 110400)
-#  define MRN_HAVE_SET_OPTIMAIZER_COSTS
 #endif
 
 #if defined(MRN_MARIADB_P) && (MYSQL_VERSION_ID >= 110800)


mrn_mysql_compat.h
```diff
@@ -906,6 +906,10 @@ using TABLE_LIST = Table_ref;
                          to))
 #endif
 
+#if defined(MRN_MARIADB_P) && (MYSQL_VERSION_ID >= 110400)
```
abetomo: The condition is the same as line 886, but is it better to separate it so that it is easier to understand?

otegami (approved):
LGTM except for the existed comment about macro definition.

ha_mroonga.cpp
```diff
@@ -5240,6 +5240,9 @@ int ha_mroonga::wrapper_open(const char* name,
     error = wrap_handler->ha_open(table, name, mode, open_options, table_def);
 #  else
     error = wrap_handler->ha_open(table, name, mode, open_options);
+#  endif
+#  ifdef MRN_HAVE_SET_OPTIMAIZER_COSTS
+    wrap_handler->set_optimizer_costs(ha_thd());
```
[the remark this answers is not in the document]
komainu8: Oh...
I didn't notice..

mrn_mysql_compat.h
komainu8: I summarized it because it was redundant.

Issue #984, opened 2025-09-22
Failed to normalize key when we use "FORCE INDEX" and "BETWEEN" on MariaDB 11.8


komainu8:
### What happened?

```
CURRENT_TEST: plugin/mroonga/storage/select.secondary_key
  mysqltest: At line 48: query 'select * from t1 force index(idx1) where c2 between 'c' and 'h' order by c1 asc' failed: ER_ERROR_ON_READ (1024): [key][normalize] failed to normalize: <t1#idx1>: <c>(1)
  
  The result from queries just before the failure was:
  < snip >
  c1	c2	c3
  3	30	sa si su se so
  6	30	aa bb cc dd ee
  drop table t1;
  create table t1 (c1 varchar(5) primary key, c2 varchar(5), c3 text, key idx1(c2), fulltext index ft(c3))engine=mroonga;
  insert into t1 values('ab','ijk',"aa ii uu ee oo");
  insert into t1 values('bc','ghi',"ka ki ku ke ko");
  insert into t1 values('cd','efg',"sa si su se so");
  insert into t1 values('de','cde',"ta ti tu te to");
  insert into t1 values('ef','abc',"aa ii uu ee oo");
  select * from t1 force index(idx1) where c2 < 'e' order by c1 asc;
  c1	c2	c3
  de	cde	ta ti tu te to
  ef	abc	aa ii uu ee oo
  select * from t1 force index(idx1) where c2 > 'e' order by c1 asc;
  c1	c2	c3
  ab	ijk	aa ii uu ee oo
  bc	ghi	ka ki ku ke ko
  cd	efg	sa si su se so
  select * from t1 force index(idx1) where c2 between 'c' and 'h' order by c1 asc;
  
  More results from queries before failure can be found in /usr/share/mariadb-test/var/2/log/secondary_key.log
  
  Warnings from just before the error:
  Error 1024 [key][normalize] failed to normalize: <t1#idx1>: <c>(1)
```

### How to reproduce it

```sql
drop table if exists t1, t2, t3;

create table t1 (c1 int primary key, c2 int, c3 text, key idx1(c2), fulltext index ft(c3));
insert into t1 values(1,10,"aa ii uu ee oo");
insert into t1 values(2,20,"ka ki ku ke ko");
insert into t1 values(3,30,"sa si su se so");
insert into t1 values(4,40,"ta ti tu te to");
insert into t1 values(5,50,"aa ii uu ee oo");

select * from t1;
[...]

PR #999, opened 2025-11-10
Setting encoding when we convert between() function

komainu8:
Fix: GH-984

This issue occurs because normalization fails inside groonga-normalizer-mysql.
It happens when the `between()` function is condition-pushed down.

When the `between()` function is condition-pushed down, Mroonga converts it into an internal representation so it can be processed internally.
However, during this conversion, the encoding is not set.
As a result, the converted `between()` function with the encoding `GRN_ENC_NONE`.

Normalization in groonga-normalizer-mysql assumes that the encoding is `GRN_ENC_UTF8`.
Therefore, when the encoding is `GRN_ENC_NONE`, normalization fails.

When condition push down occurs, we can fix this issue by applying that encoding information inside Mroonga as well.
The encoding setting is provided by MariaDB.

### DIFF
--- a/lib/mrn_condition_converter.cpp
+++ b/lib/mrn_condition_converter.cpp
@@ -981,10 +981,10 @@ namespace mrn {
     const Item_field* field_item = static_cast<const Item_field*>(target_item);
     append_field_value(field_item, expression);
 
-    enum_field_types field_type = field_item->field->real_type();
-    NormalizedType normalized_type = normalize_field_type(field_type);
+    auto field_type = field_item->field->real_type();
+    auto normalized_type = normalize_field_type(field_type);
     if (normalized_type == STRING_TYPE) {
-      grn_encoding encoding = encoding::convert(field_item->field->charset());
+      auto encoding = encoding::convert(field_item->field->charset());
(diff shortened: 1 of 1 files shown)


lib/mrn_condition_converter.cpp
```diff
@@ -980,6 +981,16 @@ namespace mrn {
     const Item_field* field_item = static_cast<const Item_field*>(target_item);
     append_field_value(field_item, expression);
 
+    enum_field_types field_type = field_item->field->real_type();
+    NormalizedType normalized_type = normalize_field_type(field_type);
```
kou:
```suggestion
    auto normalized_type = normalize_field_type(field_type);
```

lib/mrn_condition_converter.cpp
```diff
@@ -980,6 +981,16 @@ namespace mrn {
...
+    if (normalized_type == STRING_TYPE) {
+      grn_encoding encoding = encoding::convert(field_item->field->charset());
```
kou:
```suggestion
      auto encoding = encoding::convert(field_item->field->charset());
```

abetomo (approved):

PR #1005, opened 2025-11-27
test wrapper: use "tokenizer" instead of "parser"

komainu8:
Because 'parser' is deprecated and will be removed in a future release.


komainu8:
Ah...
This is meaningless.

Issue #1033, opened 2026-01-11
wrapper: Drop support for `enigne "XXX"` in `CONNECTION`


kou:
### What is your problem?

I didn't notice it but Mroonga accepts `engine "XXX"` for wrapper mode in not only `COMMENT` but also `CONNECTION`: https://jira.mariadb.org/browse/MDEV-38530

It's not documented and tested. Let's drop support for the feature.

I haven't tried this but this may be able to remove the code for `CONNECTION` support.

```diff
diff --git a/mrn_table.cpp b/mrn_table.cpp
index bb430c11a..0700cb879 100644
--- a/mrn_table.cpp
+++ b/mrn_table.cpp
@@ -348,9 +348,9 @@ int mrn_parse_table_param(MRN_SHARE *share, TABLE *table)
     &part_elem, &sub_elem);
 #endif
 #ifdef WITH_PARTITION_STORAGE_ENGINE
-  for (i = 4; i > 0; i--)
+  for (i = 3; i > 0; i--)
 #else
-  for (i = 2; i > 0; i--)
+  for (i = 1; i > 0; i--)
 #endif
   {
     const char *params_string_value = NULL;
@@ -358,7 +358,7 @@ int mrn_parse_table_param(MRN_SHARE *share, TABLE *table)
     switch (i)
     {
 #ifdef WITH_PARTITION_STORAGE_ENGINE
-      case 4:
+      case 3:
         if (!sub_elem || !sub_elem->part_comment)
           continue;
         DBUG_PRINT("info", ("mroonga create sub comment string"));
@@ -367,7 +367,7 @@ int mrn_parse_table_param(MRN_SHARE *share, TABLE *table)
         DBUG_PRINT("info",
                    ("mroonga sub comment string=%s", params_string_value));
         break;
-      case 3:
+      case 2:
         if (!part_elem || !part_elem->part_comment)
           continue;
         DBUG_PRINT("info", ("mroonga create part comment string"));
@@ -377,7 +377,7 @@ int mrn_parse_table_param(MRN_SHARE *share, TABLE *table)
                    ("mroonga part comment string=%s", params_string_value));
         break;
 #endif
-      case 2:
+      default:
         if (LEX_STRING_IS_EMPTY(table->s->comment))
           continue;
         DBUG_PRINT("info", ("mroonga create comment string"));
@@ -387,16 +387,6 @@ int mrn_parse_table_param(MRN_SHARE *share, TABLE *table)
[...]

PR #1036, opened 2026-01-19
Drop support for CONNECTION comment

komainu8:
Fix: GH-1033.


komainu8:
Already included in commit 6279f7afeef5a015d03a24eebcb612df6df606c1, so this is not needled in this PR.

PR #1041, opened 2026-02-03
Fix a crash when mroonga_default_tokenizer is set to NULL

hadeer-r:
- This PR fixes a SIGSEGV crash occurring when the system variable mroonga_default_tokenizer is set to NULL.
- The crash was caused by passing a NULL pointer to strcmp() in the update function of the mroonga_default_tokenizer variable.

**Changes**
- Mapped NULL input to "off" behavior.
- Included a test case to verify the fix.

This is the same fix I submitted to MariaDB here: https://github.com/MariaDB/server/pull/4606

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -1007,6 +1007,9 @@ static void mrn_default_tokenizer_update(THD* thd,
 
   if (!new_value) {
     new_value = "off";
+#ifndef MRN_NEED_FREE_STRING_MEMALLOC_PLUGIN_VAR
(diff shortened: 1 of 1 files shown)


komainu8:
The following CI is currently failing, but this is unrelated to this change. I will fix it in a separate update.

- https://github.com/mroonga/mroonga/actions/runs/21666213023/job/62556945640?pr=1041
- https://github.com/mroonga/mroonga/actions/runs/21666213023/job/62556945647?pr=1041

Issue #1063, opened 2026-02-23
Interrupting mysqldump before reaching Mroonga table will crash mysqld


fdiary:
### What happened?

I did `mysqldump ... | less` for the database containing mroonga tables then typed 'q' immediately, then mysqld crashed with the following error log.

```
260223 14:02:27 [ERROR] /.../mysqld got signal 11 ;
Sorry, we probably made a mistake, and this is a bug.
...
Attempting backtrace. Include this in the bug report.
(note: Retrieving this information may fail)

Thread pointer: 0x7f1d2c000c58
stack_bottom = 0x7f1d543ed000 thread_stack 0x49000
mysys/stacktrace.c:216(my_print_stacktrace)[0x55fe8e7e27ce]
sql/signal_handler.cc:230(handle_fatal_signal)[0x55fe8e29374d]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x13140)[0x7f1d8f91d140]
mroonga-15.25/ha_mroonga.cpp:14134(ha_mroonga::generic_reset())[0x7f1d8c8b3775]
mroonga-15.25/ha_mroonga.cpp:14196(ha_mroonga::reset())[0x7f1d8c8b3975]
sql/sql_base.cc:1074(close_thread_table(THD*, TABLE**))[0x55fe8df8a5c3]
sql/sql_base.cc:1006(close_thread_tables(THD*))[0x55fe8df8a8f3]
sql/sql_base.cc:2732(Locked_tables_list::unlock_locked_tables(THD*))[0x55fe8df8b81b]
sql/sql_class.cc:1673(THD::cleanup())[0x55fe8dfa6bd0]
psi/mysql_thread.h:745(unlink_thd(THD*))[0x55fe8def0205]
sql/sql_connect.cc:1516(do_handle_one_connection(CONNECT*, bool))[0x55fe8e13fbd1]
sql/sql_connect.cc:1416(handle_one_connection)[0x55fe8e14018d]
perfschema/pfs.cc:2201(pfs_spawn_thread)[0x55fe8e4d848e]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x7ea7)[0x7f1d8f911ea7]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x3f)[0x7f1d8f4f8acf]

Connection ID (thread ID): 17
Status: KILL_CONNECTION
Query (0x0): (null)
```

[...]

PR #1066, opened 2026-03-04
Fix crash in generic_reset() by tracking st_mrn_ft_info creation

abetomo:
GitHub fixes GH-1063

When a connection is interrupted after a query completes, cleanup could be called on already freed resources, causing a crash.
By tracking whether `st_mrn_ft_info` has been created, we skip the cleanup in generic_reset() if it has already been done.

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -3101,6 +3101,7 @@ ha_mroonga::ha_mroonga(handlerton* hton, TABLE_SHARE* share_arg)
       ignoring_duplicated_key(false),
       inserting_with_update(false),
       fulltext_searching(false),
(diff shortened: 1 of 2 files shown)


ha_mroonga.cpp
```diff
@@ -10165,6 +10166,8 @@ ha_mroonga::generic_ft_init_ext_select(uint flags, uint key_nr, String* key)
   info->id_accessor = NULL;
   info->key_accessor = NULL;
 
+  st_mrn_ft_infos_created_ = true;
```
kou: Is not `generic_ft_init_ext_select()` nested? If it's nested, this will not work.

abetomo:
Remove generic_reset().
#1068

PR #1069, opened 2026-03-13
Fix SIGSEGV on NULL mroonga_log_file

hadeer-r:
- Fix SIGSEGV when mroonga_log_file is set to NULL or empty string
- Add mroonga_log_file_check() to reject NULL/empty values
- Add MTR test to cover this case

this same fix I submitted to MariaDB: https://github.com/MariaDB/server/pull/4801

### DIFF
--- a/ha_mroonga.cpp
+++ b/ha_mroonga.cpp
@@ -887,6 +887,7 @@ static int mrn_log_file_check(THD* thd,
 
   DBUG_RETURN(0);
 }
+
 static void
 mrn_log_file_update(THD* thd, mrn_sys_var* var, void* var_ptr, const void* save)
 {


ha_mroonga.cpp
```diff
@@ -864,6 +864,29 @@ static MYSQL_SYSVAR_ENUM(log_level,
...
+
+  if (!new_value || len == 0) {
+    my_printf_error(ER_MRN_INVALID_NULL_VALUE_NUM,
+                    ER_MRN_INVALID_NULL_VALUE_STR,
+                    MYF(0),
+                    "mroonga_log_file");
+    DBUG_RETURN(1);
+  }
+
+  *((const char**)save) = new_value;
```
abetomo: Thanks for the PR!

`new_value` is a pointer to `buf`, which is only valid within `mrn_log_file_check()`.
Saving it to `save` means it may become an invalid address outside the function.

MySQL and MariaDB use `thd->strmake()` for this.
https://github.com/mysql/mysql-server/blob/447eb26e094b444a88c532028647e48228c3c04f/sql/sql_plugin_var.cc#L672-L682
https://github.com/MariaDB/server/blob/ef4be39bfcbae1b0090a3098e511d14457d6139f/sql/sql_plugin.cc#L2801-L2813

Could you set `save` in the same way?

hadeer-r: Thank you for the review! I've updated it to use `thd->strmake()`.

abetomo (approved):
Thanks for the fixes!

ha_mroonga.cpp
```diff
@@ -864,6 +864,29 @@ static MYSQL_SYSVAR_ENUM(log_level,
...
+                    MYF(0),
+                    "mroonga_log_file");
+    DBUG_RETURN(1);
+  }
+
+  *((const char**)save) = thd->strmake(new_value, len);
+
+  DBUG_RETURN(0);
+}
 static void
```
kou: An empty line is missing.

```suggestion

static void
```

PR #1116, opened 2026-07-02
package percona-server-8.4: Don't enable PGO

kou:
If we enable PGO (Profile-Guided Optimization), Percona Server will be built twice. It causes a build error because we don't provide `mysql-test/var` directory that needs PGO enabled build.

https://github.com/percona/percona-server/blob/aa9cdab9f209d9e60886a1320b2d7111bb796a56/build-ps/percona-server.spec#L690-L761

Mroonga doesn't need full Percona Server build. So our build system builds only needed parts to reduce build time. PGO enabled build is also not needed for us.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -181,14 +181,12 @@ jobs:
             test-image: "images:almalinux/9"
 
           # Percona Server 8.4
-          # This is a temporary workaround.
-          # The build fails due to what appears to be an upstream issue, so we skip it for now.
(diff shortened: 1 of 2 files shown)


komainu8:
Could you enable https://github.com/mroonga/mroonga/blob/main/.github/workflows/linux.yml#L183-L191 and run the CI, just to be safe?

kou:
Oh, I missed it.

komainu8 (approved):

PR #1136, opened 2026-08-07
packages debian-trixie: add MySQL 9.7 LTS

komainu8:
This change adds support for MySQL 9.7 LTS on Debian 13 (Trixie).

However, the Mroonga Debian package cannot be installed with MySQL 9.7 because of a dependency issue.
For details, see: https://bugs.mysql.com/bug.php?id=120697.

Therefore, we do not provide the package now.
 This change only adds the necessary definitions.


komainu8:
Do you have any comments?

